Issue #28, opened 2022-06-30
Add instructions on how to use Dredd


afd:
Under an `examples` directory, add to the Dredd repo:

* A single C++ file, showing how Dredd can be invoked without a compilation database

* A simple C++ project, to show how CMake can be used to generate a compilation database for the project and then how Dredd can be applied to the project.

We may as well update the Dredd CI to follow these commands, and we could also have the CI check that Dredd leads to very precise output for both the example and the project.

PR #32, opened 2022-07-05
Add single file example

JamesLee-Jones:
Adds a small example to illustrate how dredd works and updates README.

Part of #28.


examples/simple/pi.cc
```diff
@@ -0,0 +1,33 @@
...
+
+using namespace std;
+
+static double calculatePi() {
+    double sum = 0.0;
+    int numerator = -1;
+
+    for (int d = 1; d < 1000000; d += 2) {
+        numerator *= -1;
+        sum+= numerator / double(d);
```
afd, line 25: Nit: space before +=

afd (changes requested):
Looks good - one small nit.

examples/simple/pi.cc
JamesLee-Jones: Done

README.md
```diff
@@ -44,6 +44,19 @@ without having to build it from source.
 TODO(https://github.com/mc-imperial/dredd/issues/28): add instructions
 on how to use Dredd
 
+`pi.cc` is a small example in `examples/simple` that calculates an approximation of pi. To run this example, run the following from the root of the repository:
```
afd, line 47: I suggest starting with a preamble:

We first show how to apply Dredd to a simple stand-alone program. We then show how to apply it to a C++ CMake project.

Before trying these instructions out, make sure you have built Dredd and that the `dredd` executable is located under [the third_party clang bin dir]

### Applying Dredd to a single file example

README.md
```diff
@@ -44,6 +44,19 @@ without having to build it from source.
...
+cp examples/simple/pi.cc examples/simple/backup.cc
+third_party/clang+llvm-13.0.1/bin/dredd examples/simple/pi.cc
+clang examples/simple/pi.cc -lstdc++ -o examples/simple/pi
+```
+Running the executable with no additional arguments via `examples/simple/pi` will produce the expected result of 3.14159.  You can specify which mutants you want to use at runtime by running:
+```
+DREDD_ENABLED_MUTATION=$i$ examples/simple/pi
+```
+where $i$ is the id of the mutation in the program generated by dredd. This will print a different output to the original program and in some cases may not terminate.
+
```
afd, line 59: Now you can add:

### Applying Dredd to a CMake project

and you can move the TODO from above in here, changing the comment to be more specific - add instructions on how to apply Dredd to a larger project.

README.md
```diff
@@ -44,6 +44,19 @@ without having to build it from source.
 TODO(https://github.com/mc-imperial/dredd/issues/28): add instructions
 on how to use Dredd
 
+`pi.cc` is a small example in `examples/simple` that calculates an approximation of pi. To run this example, run the following from the root of the repository:
+```
+# This will save the example file since pi.cc is modified in-place by dredd.
+cp examples/simple/pi.cc examples/simple/backup.cc
```
afd, line 50: Rather than having them copy pi.cc into a backup, I suggest merely including the command to mutate in place, then remarking on this (possibly noting that they can use git diff to see the changes that Dredd has made, but noting that these changes are not meant to be human readable), and then at the end of the section giving them a git checkout command to restore the file.

README.md
```diff
@@ -44,6 +44,19 @@ without having to build it from source.
...
+third_party/clang+llvm-13.0.1/bin/dredd examples/simple/pi.cc
+clang examples/simple/pi.cc -lstdc++ -o examples/simple/pi
```
afd, line 52: Change `clang` to `clang++` and then you should be able to remove `-lstdc++`. You could also note in a comment that `clang++` can be replaced with your favourite C++ compiler.

It would also be good to note that the compile command may lead to warnings being issued, due to the changes Dredd has made, and that this is OK.

README.md
```diff
@@ -44,6 +44,19 @@ without having to build it from source.
...
+```
+Running the executable with no additional arguments via `examples/simple/pi` will produce the expected result of 3.14159.  You can specify which mutants you want to use at runtime by running:
+```
+DREDD_ENABLED_MUTATION=$i$ examples/simple/pi
+```
+where $i$ is the id of the mutation in the program generated by dredd. This will print a different output to the original program and in some cases may not terminate.
```
afd, line 58: Before "This will print" I would add:

For example, running ... will enable mutation number 3.

so that they have a very concrete example.

Also: "This will print a different output" - change it to "This may", and remark that some mutants might be *equivalent* mutants, in which case the value printed to the console will be unchanged.

examples/simple/pi.cc
```diff
@@ -0,0 +1,33 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <iostream>
+
+using namespace std;
+
+static double calculatePi() {
```
afd, line 19: The project uses the Google C++ style, so I suggest naming this `CalculatePi` in line with the style guide.

examples/simple/pi.cc
```diff
@@ -0,0 +1,33 @@
...
+    for (int d = 1; d < 1000000; d += 2) {
+        numerator *= -1;
+        sum += numerator / double(d);
+    }
+
+    return sum * 4;
+}
+
+int main() {
+    cout << calculatePi() << "\n";
```
afd, line 32: I suggest removing the `using std`, changing `cout` to `std::cout`, and changing "\n" to `std::endl`

examples/simple/pi.cc
```diff
@@ -0,0 +1,33 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <iostream>
+
+using namespace std;
```
afd, line 17: I would remove this (see below)

examples/simple/pi.cc
```diff
@@ -0,0 +1,33 @@
...
+    double sum = 0.0;
+    int numerator = -1;
+
+    for (int d = 1; d < 1000000; d += 2) {
+        numerator *= -1;
+        sum += numerator / double(d);
+    }
+
+    return sum * 4;
+}
```
afd, line 29: Can you use 2 space indentation (so that we're following Google C++ style)?

afd (changes requested):
Thanks for this example and description, which looks good.

I have read it properly now and suggested some improvements.

README.md
```diff
@@ -41,8 +41,37 @@ TODO(https://github.com/mc-imperial/dredd/issues/29): provide support
...
+Running the executable with no additional arguments via `examples/simple/pi` will produce the expected result of 3.14159.  You can specify which mutants you want to use at runtime by running:
+```
+DREDD_ENABLED_MUTATION=$i$ examples/simple/pi
+```
+where $i$ is the id of the mutation in the program generated by Dredd. For example, running `DREDD_ENABLE_MUTATION=2 examples/simple/pi` will enable mutation number 2 in the modified program. 
+This may print a different output to the original program and in some cases may not terminate. 
+There is a possibility that equivalent mutants are generated which can lead to the printed value being unchanged.
+
+To restore the file `pi.cc` to it's initial state, run
+`git checkout HEAD examples/simple/pi.cc`
```
afd, line 69: Nit: I guess it would also be nice to show the `rm` command to get rid of the generated executable.

afd (approved):
These look great! One small request added - once you attend to that I'll be happy to merge.

afd (approved):
LGTM - approved.

PR #37, opened 2022-07-18
Fixed pull request comments

### LINKED ISSUE #28: Add instructions on how to use Dredd

JamesLee-Jones:
Wrote a small math library and test suite to show how Dredd can be applied to a CMake project + updated README.

Fixed #28.


afd:
Let's close this and revert to the original pull request.

PR #46, opened 2022-07-21
Add single-file test framework

### LINKED ISSUE #45: Add single file test cases to allow regression testing
Add a script that can be pointed at a directory of single-file test cases and will check that each test case ends up being identical to some expected output after mutation.

This will allow us to add regression tests in response to bugs that we find in Dredd.

afd:
Adds a script and corresponding initial tests to allow confirming that
Dredd performs as expected on a number of single-file test cases.
    
Fixes #45.


JamesLee-Jones (approved):
These changes look good!

PR #47, opened 2022-07-21
Added support for floating point operators

### LINKED ISSUE #38: Support floating point operators
Currently operator mutation only supports integer operators. It would be good to support floating point types too.

JamesLee-Jones:
Fixes #38


src/libdredd/src/mutate_visitor.cc
```diff
@@ -99,17 +99,20 @@ bool MutateVisitor::VisitBinaryOperator(
   // In particular, we do not want to mess with pointer arithmetic.
   const auto* result_type =
       binary_operator->getType()->getAs<clang::BuiltinType>();
-  if (result_type == nullptr || !result_type->isInteger()) {
+  if (result_type == nullptr ||
+      !(result_type->isInteger() || result_type->isFloatingPoint())) {
```
afd, line 103: Now that this check has gotten a bit more complicated, I think there's a case for extracting out the logic into a private helper method, `IsTypeSupported`.

This should be prototyped in the class, in `mutate_visitor.h`, and implemented in `mutate_visitor.cc`.

src/libdredd/src/mutate_visitor.cc
```diff
@@ -99,17 +99,20 @@ bool MutateVisitor::VisitBinaryOperator(
...
     return true;
   }
   const auto* lhs_type =
       binary_operator->getLHS()->getType()->getAs<clang::BuiltinType>();
-  if (lhs_type == nullptr || !lhs_type->isInteger()) {
+  if (lhs_type == nullptr ||
+      !(lhs_type->isInteger() || lhs_type->isFloatingPoint())) {
```
afd, line 109: Call the helper.

src/libdredd/src/mutate_visitor.cc
```diff
@@ -99,17 +99,20 @@ bool MutateVisitor::VisitBinaryOperator(
...
     return true;
   }
   const auto* rhs_type =
       binary_operator->getRHS()->getType()->getAs<clang::BuiltinType>();
-  if (rhs_type == nullptr || !rhs_type->isInteger()) {
+  if (rhs_type == nullptr ||
+      !(rhs_type->isInteger() || rhs_type->isFloatingPoint())) {
```
afd, line 115: Call the helper.

afd (changes requested):
The code changes look good! One comment about refactoring.

However, this change needs some tests.

Can you please:

- Pull main and rebase over my latest changes
- Add a couple of tests to `mutate_binary_operator_test.cc`, along the lines of the ones that are there, but for floating-point operators; e.g. `/` on `float` and `-=` on `double` (rather arbitrarily)
- Add a small synthetic example that does a few floating-point operations under `test/single_file`, with a corresponding expectation file. See the README under `test` for how to generate the expectation file automatically - but do check that it looks right before you commit it.

src/libdredd/src/mutate_visitor.cc
JamesLee-Jones: Done

src/libdredd/src/mutate_visitor.cc
JamesLee-Jones: Done

src/libdredd/src/mutate_visitor.cc
JamesLee-Jones: Done

src/libdreddtest/src/mutation_replace_binary_operator_test.cc
```diff
@@ -201,5 +201,68 @@ void foo() {
...
+    case 0: return arg1() += arg2();
+    case 1: return arg1() = arg2();
+    case 2: return arg1() /= arg2();
+    case 3: return arg1() *= arg2();
+    default: return arg1() -= arg2();
+  }
+}
+
+)";
+  const int num_replacements = 4;
```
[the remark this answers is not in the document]
afd: Similar comment for other uses of num_replacements above.

src/libdredd/src/mutate_visitor.cc
```diff
@@ -99,17 +103,17 @@ bool MutateVisitor::VisitBinaryOperator(
   // In particular, we do not want to mess with pointer arithmetic.
   const auto* result_type =
       binary_operator->getType()->getAs<clang::BuiltinType>();
-  if (result_type == nullptr || !result_type->isInteger()) {
+  if (IsTypeSupported(result_type)) {
```
afd, line 106: Given that `result_type`, `lhs_type` and `rhs_type` are never used again, it would be better to change `IsTypeSupported` so that it accepts a `QualType` as an argument, and then you can just do:

`if (IsTypeSupported(binary_operator->getType())`

and similar below with `getType()` on the LHS and RHS.

The `getAs<clang::BuiltinType>()` can then go inside the `IsTypeSupported` method.

afd (changes requested):
This is getting there - see latest batch of comments.

src/libdredd/src/mutate_visitor.cc
```diff
@@ -37,6 +37,11 @@ MutateVisitor::MutateVisitor(const clang::CompilerInstance& compiler_instance)
     : compiler_instance_(compiler_instance),
       first_decl_in_source_file_(nullptr) {}
 
+bool MutateVisitor::IsTypeSupported(const clang::QualType qual_type) {
+  const auto* type = qual_type->getAs<clang::BuiltinType>();
```
afd, line 41: nit: I would call the variable `builtin_type`, to better distinguish it from `qual_type`.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -50,6 +50,19 @@ MutationReplaceBinaryOperator::MutationReplaceBinaryOperator(
     const clang::BinaryOperator& binary_operator)
     : binary_operator_(binary_operator) {}
 
+bool MutationReplaceBinaryOperator::IsValidReplacementOperator(
+    clang::BinaryOperatorKind op) const {
+  const auto* lhs_type =
+      binary_operator_.getLHS()->getType()->getAs<clang::BuiltinType>();
+  const auto* rhs_type =
+      binary_operator_.getRHS()->getType()->getAs<clang::BuiltinType>();
+  return !((lhs_type->isFloatingPoint() || rhs_type->isFloatingPoint()) &&
```
afd, line 59: I would:

- add assertions that both lhs_type and rhs_type are either integer or floating-point (this should be true by construction, but it's good to confirm); use the && trick to have a suitable string associated with the assertions.
- add a comment noting that certain binary operators are not compatible with floating-point arguments.

afd (changes requested):
This looks great! I have noted a couple of small things, and would normally approve and let you do them and then merge. However, since you're going to have to rework your changes slightly once I land my pending PR, I've hit "request changes" so that I can take a quick look at your revision in due course.

afd:
@JamesLee-Jones Ah, the CI already fails because I made a change yesterday that uses if statements instead of switch statements when checking which binary operator mutation to apply. I suggest you hold off until I land my pending change about boolean arrays, and then you can fix things up.

src/libdredd/src/mutate_visitor.cc
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Done.

afd (approved):
Looks great!

Issue #55, opened 2022-07-26
Support enabling multiple mutants simultaneously


afd:
- [ ] Replace calls to the function that checks which mutant is enabled with a call to a predicate that returns whether a particular mutant is enabled. This will involve changing switch statements to if statements in the case of binary operator mutations.

- [ ] Implement the function efficiently using a bitset, based on the fact that for a given file we know the range of mutants that occur in the file, so we can have a bitset array just large enough to allow the relevant queries.

PR #57, opened 2022-07-27
Use bool array to track which mutants are enabled

afd:
Paving the way for having multiple mutants enabled, this change
introduces an array of booleans to track the mutations that are
enabled on a per file basis. At present, at most one element of this
array will be set to true. In a future change, environment variable
parsing will be updated to support multiple mutants.

Part of #55.

### DIFF
--- a/src/libdredd/src/mutation_remove_statement.cc
+++ b/src/libdredd/src/mutation_remove_statement.cc
@@ -34,7 +34,7 @@ MutationRemoveStatement::MutationRemoveStatement(const clang::Stmt& statement)
 
 void MutationRemoveStatement::Apply(
     clang::ASTContext& ast_context, const clang::Preprocessor& preprocessor,
-    int& mutation_id, clang::Rewriter& rewriter,
+    int first_mutation_id_in_file, int& mutation_id, clang::Rewriter& rewriter,
     std::unordered_set<std::string>& dredd_declarations) const {
   (void)dredd_declarations;  // Unused
   clang::CharSourceRange source_range = clang::tooling::maybeExtendRange(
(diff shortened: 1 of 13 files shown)


src/libdredd/src/mutation_remove_statement.cc
```diff
@@ -43,9 +43,10 @@ void MutationRemoveStatement::Apply(
       clang::tok::TokenKind::semi, ast_context);
 
   bool result = rewriter.ReplaceText(
-      source_range, "if (!__dredd_enabled_mutation(" +
-                        std::to_string(mutation_id) + ")) { " +
-                        rewriter.getRewrittenText(source_range) + " }");
+      source_range,
+      "if (!__dredd_enabled_mutation(" +
+          std::to_string(mutation_id - first_mutation_id_in_file) + ")) { " +
```
JamesLee-Jones, line 48: It may be worth mentioning again that this is to convert to the file specific mutation ID.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -269,7 +269,8 @@ void MutationReplaceBinaryOperator::Apply(
           rewriter.getRewrittenText(lhs_source_range_in_main_file) +
           +"); }, [&]() -> " + rhs_type + " { return static_cast<" + rhs_type +
           ">(" + rewriter.getRewrittenText(rhs_source_range_in_main_file) +
-          "); }, " + std::to_string(mutation_id) + ")");
+          "); }, " + std::to_string(mutation_id - first_mutation_id_in_file) +
```
JamesLee-Jones, line 272: Same comment as above.

JamesLee-Jones (approved):
Other than a couple of small comments, it's pretty clear what is happening so shouldn't require much more explanation.

src/libdredd/src/mutation_remove_statement.cc
afd: Done.

src/libdredd/src/mutation_replace_binary_operator.cc
afd: Done.

PR #59, opened 2022-07-27
Support comma-separated list of enabled mutants

### LINKED ISSUE #55: Support enabling multiple mutants simultaneously

afd:
Allows multiple mutants to be enabled simultaneously via a
comma-separated list.
    
Fixes #55.


src/libdredd/src/mutate_ast_consumer.cc
```diff
@@ -114,24 +115,46 @@ void MutateAstConsumer::HandleTranslationUnit(clang::ASTContext& context) {
...
+  // The environment variable for mutations is set, so process the contents of
+  // this environment variable as a comma-seprated list of strings.
+  dredd_prelude << "      std::string contents(dredd_environment_variable);\n";
+  dredd_prelude << "      while (true) {\n";
+  // Find the position of the next comma.
+  dredd_prelude << "        size_t pos = contents.find(\",\");\n";
+  // The next token is either the prefix before the next comma (if there is a
+  // comma) or the whole string (if there is no comma).
+  dredd_prelude << "        std::string token = (pos == std::string::npos ? "
+                   "contents : contents.substr(0, pos));\n";
```
JamesLee-Jones, line 130: It may be worth reversing the order of the comment to mention the whole string first, then the prefix before the next comma just so the order matches the conditional.

JamesLee-Jones (approved):
All looks good! Just one small change to make a comment more intuitive then it will be ready to merge.

src/libdredd/src/mutate_ast_consumer.cc
afd: Good suggestion - done!

PR #61, opened 2022-08-01
Run InstCombine tests in CI

### LINKED ISSUE #40: Run LLVM tests in Dredd CI
Currently the Dredd CI builds an LLVM library after mutation (I think it's InstCombine). It would be good if the CI could also use llvm-lit to run the InstCombine unit tests.

JamesLee-Jones:
Run tests for the library InstCombine which is build on the CI to check
that the mutations applied don't affect functionality.

Fixes #40.


JamesLee-Jones:
It turned out to be impractical to run the InstCombine tests in the CI. This has instead been replaced by #83.

PR #62, opened 2022-08-02
Refactor binary function name generation

JamesLee-Jones:
Separate binary function name generation into it's own function to 
make the Apply function easier to understand.

### DIFF
--- a/src/libdredd/src/mutation_replace_binary_operator.cc
+++ b/src/libdredd/src/mutation_replace_binary_operator.cc
@@ -71,130 +71,132 @@ bool MutationReplaceBinaryOperator::IsValidReplacementOperator(
 
 std::string MutationReplaceBinaryOperator::GetFunctionName(
     clang::ASTContext& ast_context) const {
-  std::string lhs_type = binary_operator_.getLHS()
-                             ->getType()
-                             ->getAs<clang::BuiltinType>()
-                             ->getName(ast_context.getPrintingPolicy())
-                             .str();
-  std::string rhs_type = binary_operator_.getRHS()
-                             ->getType()
-                             ->getAs<clang::BuiltinType>()
-                             ->getName(ast_context.getPrintingPolicy())
-                             .str();
-
-  if (binary_operator_.isAssignmentOp()) {
-    clang::QualType qualified_lhs_type = binary_operator_.getLHS()->getType();
-    if (qualified_lhs_type.isVolatileQualified()) {
-      assert(binary_operator_.getType().isVolatileQualified() &&
-             "Expected expression to be volatile-qualified since LHS is.");
-      lhs_type = "volatile " + lhs_type;
-    }
-  }
-  std::string function_name = "__dredd_replace_binary_operator_";
+  std::string result = "__dredd_replace_binary_operator_";
 
   // A string corresponding to the binary operator forms part of the name of the
   // mutation function, to differentiate mutation functions for different
   // operators
   switch (binary_operator_.getOpcode()) {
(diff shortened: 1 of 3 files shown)


src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -69,171 +69,203 @@ bool MutationReplaceBinaryOperator::IsValidReplacementOperator(
...
+  std::string lhs_type = binary_operator_.getLHS()
+                             ->getType()
+                             ->getAs<clang::BuiltinType>()
+                             ->getName(ast_context.getPrintingPolicy())
+                             .str();
+  std::string rhs_type = binary_operator_.getRHS()
+                             ->getType()
+                             ->getAs<clang::BuiltinType>()
+                             ->getName(ast_context.getPrintingPolicy())
+                             .str();
```
afd, line 83: Now that we don't need to use (these versions of) lhs_type and rhs_type elsewhere, I think it would make sense simply to build `function_name` up bit by bit. So first populate it with `__dredd_replace_binary_operator_` and then the operator name, and after that tack on the LHS and RHS types, and call `SpaceToUnderscore` on the fly as you tack them on.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -69,171 +69,203 @@ bool MutationReplaceBinaryOperator::IsValidReplacementOperator(
...
-
-  return new_function.str();
-}
-
-void MutationReplaceBinaryOperator::Apply(
-    clang::ASTContext& ast_context, const clang::Preprocessor& preprocessor,
-    int first_mutation_id_in_file, int& mutation_id, clang::Rewriter& rewriter,
-    std::unordered_set<std::string>& dredd_declarations) const {
-  std::string new_function_name = "__dredd_replace_binary_operator_";
+  std::string function_name = "__dredd_replace_binary_operator_";
```
afd, line 93: A good convention is to call the thing you will ultimately return `result`, hence I would rename this to `result`.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -69,171 +69,203 @@ bool MutationReplaceBinaryOperator::IsValidReplacementOperator(
...
       assert(false && "Unsupported opcode");
   }
 
+  // To avoid problems of ambiguous function calls, the argument types (ignoring
+  // whether they are references or not) are baked into the mutation function
+  // name. Some type names have space in them (e.g. 'unsigned int'); such spaces
+  // are replaced with underscores.
+  function_name +=
+      "_" + SpaceToUnderscore(lhs_type) + "_" + SpaceToUnderscore(rhs_type);
+
```
afd, line 196: My above point is that I would move the logic for stringifying lhs and rhs types to here and just do it as part of tacking them on to the function name.

afd (approved):
A suggested simplification, but otherwise good to go.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Renamed.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Done.

PR #67, opened 2022-08-03
Fix rewriting error before first declaration

### LINKED ISSUE #66: Dredd fails when first declaration in source file starts with a macro
There is a bug where if the first declaration in the main source file looks like:

```
SOME_MACRO foo() {
  ...
}
```

then Dredd will fail when trying to insert its prelude before this declaration.

afd:
Fixes Dredd so that the source range in the main file is obtained for
the first declaration occurring in that file, before inserting Dredd's
prelude. This avoids a problem where if the first declaration starts
with a macro expansion, its begin location is not valid for an
insertion.

Fixes #66.


JamesLee-Jones (approved):
Looks good, LGTM!

PR #69, opened 2022-08-04
Do not mutate switch case expressions

### LINKED ISSUE #68: Dredd mutates negative switch cases.
In e.g. this:

```
int main() {
  switch(0) {
  case -1:
    break;
  default:
    break;
  }
}
```

Dredd treats `-1` as a unary operator and mutates it. This leads to invalid code since switch cases need to be constant.

@JamesLee-Jones for info.

afd:
Switch case expressions need to be constants and so cannot easily
mutated in the Dredd style.

Fixes #68.


JamesLee-Jones (approved):
LGTM!

PR #71, opened 2022-08-04
Correct where Dredd prelude is inserted

### LINKED ISSUE #70: Dredd prelude is inserted after typedef keyword when file starts with a typedef
When a file's first declaration is a typedef, e.g.:

```
typedef struct S {
  ...
```

the Dredd prelude gets inserted *after* the `typedef` keyword, when it should be inserted before.

afd:
Changes the logic for deciding where to insert the Dredd prelude to
account for the fact that the first declaration encountered when
visiting the AST may not correspond to the declaration whose start
appears syntactically earliest in the main source file.

Fixes #70.

### DIFF
--- a/src/libdredd/src/mutate_visitor.cc
+++ b/src/libdredd/src/mutate_visitor.cc
@@ -25,6 +25,7 @@
 #include "clang/AST/Stmt.h"
 #include "clang/AST/Type.h"
 #include "clang/Basic/SourceLocation.h"
+#include "clang/Basic/SourceManager.h"
 #include "clang/Frontend/CompilerInstance.h"
 #include "libdredd/mutation_remove_statement.h"
 #include "libdredd/mutation_replace_binary_operator.h"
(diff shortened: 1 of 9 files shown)


src/libdredd/src/mutate_visitor.cc
```diff
@@ -55,17 +55,29 @@ bool MutateVisitor::TraverseDecl(clang::Decl* decl) {
...
-    // before it.
-    first_decl_in_source_file_ = decl;
+    // been encountered so far: record it so that the dredd prelude can be
+    // inserted before it.
+    //
+    // The order in which declarations appear in the source file may not exactly
+    // match the order they are visited in the AST (for example, a typedef
+    // declaration is visited after the associated type declaration, even though
+    // the 'typedef' keyword occurs first in the AST), thus this location is
+    // updated each a declaration that appears earlier is encountered.
```
JamesLee-Jones, line 78: Should probably say "each time a declaration that appears earlier is encountered."

JamesLee-Jones (approved):
Other than one small grammatical error, it looks good! Feel free to merge provided the tests pass.

src/libdredd/src/mutate_visitor.cc
afd: Thanks!

PR #74, opened 2022-08-04
Do not mutate operators that use bit-field l-values

### LINKED ISSUE #73: Avoid creating reference to bitfield
In an example like:

```
struct S {
  int a : 3;
  int b : 3;
};

void foo() {
   S myS;
   myS.a = myS.b;
}
```

Dredd will try to pass `myS.a` by reference.

This is illegal because `myS.a` is a bitfield.

Unary and binary operations that take bitfields as l-values should be avoided.

afd:
As it is not possible to pass bit-fields by refernce, mutations cannot
be applied to bit-field l-values.

Fixes #73.


JamesLee-Jones (approved):
All looks good, LGTM.

PR #75, opened 2022-08-04
Account for comments when removing return statements

### LINKED ISSUE #72: Problem with statement deletion and semicolons when there is a comment between end of statement and semicolon
Dredd does the wrong thing with this:

```
int f()
{
    return 1 /* something */ ;
}
```

when "deleting" the return statement (by turning it into a conditional).

There is logic to check whether a statement is immediately followed by a semi-colon, but it looks like that logic does not skip over comments.

As a result we end up with:

```
    if (!__dredd_enabled_mutation(0)) { return 1 } /* something */ ;
```

which is wrong.

afd:
Ensures that comments are accounted for when extending statement
source ranges to include trailing semi-colons during statement
deletion.

Fixes #72.


JamesLee-Jones (approved):
Looks good!

PR #79, opened 2022-08-09
Allow mutation inside initializers

### LINKED ISSUE #50: Mutations not being applied to variable initializers
When there is a binary operator in the initializer of a variable in a function, the operator is not mutated. This is because Dredd required the closest enclosing deceleration to be a function.

```
void foo() {
  float z = 5.346 * 3.64;
}
```

afd:
Refines the logic for deciding whether the mutation visitor is inside
a function so that expressions under variable declarations that are
themselves in functions are regarded as being in functions. This
allows mutating variable initializers inside functions, but still
disallows mutating the initializers of e.g. class members.

Care is also taken to avoid mutating constexpr initializers.

Fixes #50.


test/single_file/initializers_outside_functions.cc.expected
```diff
@@ -0,0 +1,14 @@
...
+
+  class A {
+    int z = y + y;
+    void foo() {
+      class B {
+        int w = !x;
+      };
+    }
+  };
+}
```
JamesLee-Jones, line 14: May be missing it's respective .cc file for testing.

test/single_file/initializers_outside_functions.expected
```diff
@@ -0,0 +1,14 @@
...
+
+  class A {
+    int z = y + y;
+    void foo() {
+      class B {
+        int w = !x;
+      };
+    }
+  };
+}
```
JamesLee-Jones, line 14: The naming of this file appears to be wrong, it may also be missing the relevant .cc/.c file for testing.

test/single_file/initializers_outside_functions.cc.expected
afd: Fixed, thanks!

test/single_file/initializers_outside_functions.expected
afd: Fixed, thanks!

JamesLee-Jones (approved):
Looks good!

PR #81, opened 2022-08-09
Do not mutate template arguments

### LINKED ISSUE #80: Do not mutate template instantiation expressions
In something like:

```
template<int N> void foo() {
  int A[N] = {0};
}

int main() {
  foo<1 + 2>();
}
```

the `1 + 2` should not be mutated.

afd:
Avoids mutating template argument expressions, which normally need to
be compile-time constants.

Fixes #80.


JamesLee-Jones (approved):
Looks good! LGTM

PR #82, opened 2022-08-10
Optimization for the case when no mutants are enabled

afd:
If no mutants are enabled, it makes sense to quickly return the
original expression. As this will be common when several files have
been mutated but only certain mutants are enabled, it is worth
optimising for. Benchmarking showed substantial speedups running zlib
after mutating all compression files, and then running with no mutants
enabled.


JamesLee-Jones (approved):
LGTM!

Issue #84, opened 2022-08-10
Finish planned set of Dredd mutations in a unified way


afd:
Issues #63 and #64 represent the final mutation operators planned for the initial version of Dredd. This issue sets out a proposal for how to implement them in a unified way.

Currently we have two expression-related mutation operator classes: MutationReplaceUnaryOperator and MutationReplaceBinaryOperator.

Rather than adding new classes for the operators of issues #63 and #64, I propose we proceed as follows:

* Enhance MutationReplaceUnaryOperator so that it also considers unary operator insertion (as per #63), and replacement of the whole expression with a constant (as per #64). (Aside: redundant mutations can be nicely avoided in this process: e.g. with a unary operator "!e", we don't want to have both "e" and "!!e" as replacements.)

* Similarly, enhance MutationReplaceBinaryOperator in this manner.

* Add one more class, MutationReplaceExpr, for mutating any clang::Expr that *isn't* a unary or binary operator. The constructor for this class should assert that the clang::Expr with which it is constructed is not one of these subtypes. This class should just implement unary operator insertion (as per #63) and replacement with a constant (as per #64).

In mutate visitor, the general function for visiting expressions should be overridden - I think it is VisitExpr. When it visits an expression that isn't a unary or binary operator, it should add a MutationReplaceExpr object to the set of mutations (as long as the expression has a supported result type).

afd:
@JamesLee-Jones This is the idea I mentioned today about how to do the remaining Dredd mutations in a unified way. This would be a nice coding issue for you to work on while I'm away.

Issue #64, opened 2022-08-02
Replace expression with constant


afd:
Replace R-values of integer and floating-point type with constants drawn from a limited set.

As a starting point, let us go with:

- 0 and 1 for unsigned integer types
- 0, 1 and -1 for signed integer types
- 0.0, 1.0 and -1.0 for floating-point types

It might be worth considering other special values, but it would be worth doing some reading up on that first.

Issue #63, opened 2022-08-02
Insert unary operator mutation


afd:
Add an operator that will insert certain unary operators.

The most important case for insertion if `!` on booleans. It might also be worth inserting `!` on integers, and perhaps also `~`. It is less clear that it is worth inserting `++` and `--`.

JamesLee-Jones:
This is addressed by #86.

PR #86, opened 2022-08-14
Finish planned Dredd mutations for unary operators

JamesLee-Jones:
Implement operator insertion and constant replacement for unary operators.

Part of #84, #63, #64.

### DIFF
--- a/src/libdredd/include/libdredd/mutation_replace_binary_operator.h
+++ b/src/libdredd/include/libdredd/mutation_replace_binary_operator.h
@@ -61,6 +61,7 @@ class MutationReplaceBinaryOperator : public Mutation {
   [[nodiscard]] bool IsValidReplacementOperator(
       clang::BinaryOperatorKind op) const;
 
+  // This returns a string corresponding to the non-mutated expression.
   std::string getExpr(clang::ASTContext& ast_context) const;
 
   void GenerateConstantInsertion(std::stringstream& new_function,
--- a/src/libdredd/include/libdredd/mutation_replace_binary_operator.h
+++ b/src/libdredd/include/libdredd/mutation_replace_binary_operator.h
@@ -15,6 +15,7 @@
 #ifndef LIBDREDD_MUTATION_REPLACE_BINARY_OPERATOR_H
 #define LIBDREDD_MUTATION_REPLACE_BINARY_OPERATOR_H
 
+#include <sstream>
 #include <string>
 #include <unordered_set>
 #include <vector>
@@ -22,6 +23,7 @@
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/Expr.h"
 #include "clang/AST/OperationKinds.h"
+#include "clang/AST/Type.h"
 #include "clang/Lex/Preprocessor.h"
 #include "clang/Rewrite/Core/Rewriter.h"
 #include "libdredd/mutation.h"
@@ -60,6 +62,10 @@ class MutationReplaceBinaryOperator : public Mutation {
       clang::BinaryOperatorKind op) const;
 
   std::string getExpr(clang::ASTContext& ast_context) const;
+
+  void GenerateConstantInsertion(std::stringstream& new_function,
+                                 const clang::BuiltinType* exprType,
+                                 int& mutant_offset) const;
 
   const clang::BinaryOperator& binary_operator_;
 };
--- a/src/libdredd/include/libdredd/mutation_replace_binary_operator.h
+++ b/src/libdredd/include/libdredd/mutation_replace_binary_operator.h
@@ -23,7 +23,6 @@
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/Expr.h"
 #include "clang/AST/OperationKinds.h"
-#include "clang/AST/Type.h"
 #include "clang/Lex/Preprocessor.h"
 #include "clang/Rewrite/Core/Rewriter.h"
 #include "libdredd/mutation.h"
@@ -62,11 +61,19 @@ class MutationReplaceBinaryOperator : public Mutation {
       clang::BinaryOperatorKind op) const;
 
   // This returns a string corresponding to the non-mutated expression.
-  std::string getExpr(clang::ASTContext& ast_context) const;
+  std::string GetExpr(clang::ASTContext& ast_context) const;
 
-  void GenerateConstantInsertion(std::stringstream& new_function,
-                                 const clang::BuiltinType* exprType,
-                                 int& mutant_offset) const;
+  // Replaces binary expressions with either the left or right operand.
+  void GenerateArgumentReplacement(const std::string& arg1_evaluated,
+                                   const std::string& arg2_evaluated,
+                                   std::stringstream& new_function,
+                                   int& mutant_offset) const;
+
+  // Replaces binary operators with other valid binary operators.
+  void GenerateBinaryOperatorReplacement(
+      const std::vector<clang::BinaryOperatorKind>& operators,
+      const std::string& arg1_evaluated, const std::string& arg2_evaluated,
+      std::stringstream& new_function, int& mutant_offset) const;
 
   const clang::BinaryOperator& binary_operator_;
 };
--- a/src/libdredd/include/libdredd/mutation_replace_binary_operator.h
+++ b/src/libdredd/include/libdredd/mutation_replace_binary_operator.h
@@ -59,6 +59,8 @@ class MutationReplaceBinaryOperator : public Mutation {
   [[nodiscard]] bool IsValidReplacementOperator(
       clang::BinaryOperatorKind op) const;
 
+  std::string getExpr(clang::ASTContext& ast_context) const;
+
   const clang::BinaryOperator& binary_operator_;
 };
 
--- a/src/libdredd/include/libdredd/mutation_replace_unary_operator.h
+++ b/src/libdredd/include/libdredd/mutation_replace_unary_operator.h
@@ -15,12 +15,14 @@
 #ifndef LIBDREDD_MUTATION_REPLACE_UNARY_OPERATOR_H
 #define LIBDREDD_MUTATION_REPLACE_UNARY_OPERATOR_H
 
+#include <sstream>
 #include <string>
 #include <unordered_set>
 
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/ASTFwd.h"
 #include "clang/AST/OperationKinds.h"
+#include "clang/AST/Type.h"
 #include "clang/Lex/Preprocessor.h"
 #include "clang/Rewrite/Core/Rewriter.h"
 #include "libdredd/mutation.h"
@@ -57,6 +59,10 @@ class MutationReplaceUnaryOperator : public Mutation {
   std::string getExpr(clang::ASTContext& ast_context) const;
 
   std::string GetFunctionName(clang::ASTContext& ast_context) const;
+
+  void GenerateConstantInsertion(std::stringstream& new_function,
+                                 const clang::BuiltinType* exprType,
+                                 int& mutant_offset) const;
 
   const clang::UnaryOperator& unary_operator_;
 };
--- a/src/libdredd/include/libdredd/mutation_replace_unary_operator.h
+++ b/src/libdredd/include/libdredd/mutation_replace_unary_operator.h
@@ -54,6 +54,8 @@ class MutationReplaceUnaryOperator : public Mutation {
   [[nodiscard]] bool IsValidReplacementOperator(
       clang::UnaryOperatorKind op) const;
 
+  std::string getExpr(clang::ASTContext& ast_context) const;
+
   std::string GetFunctionName(clang::ASTContext& ast_context) const;
 
   const clang::UnaryOperator& unary_operator_;
--- a/src/libdredd/include/libdredd/mutation_replace_unary_operator.h
+++ b/src/libdredd/include/libdredd/mutation_replace_unary_operator.h
@@ -47,10 +47,6 @@ class MutationReplaceUnaryOperator : public Mutation {
                                       const std::string& input_type,
                                       int& mutation_id) const;
 
-  static void ApplyCppTypeModifiers(const clang::Expr* expr, std::string& type);
-
-  static void ApplyCTypeModifiers(const clang::Expr* expr, std::string& type);
-
   [[nodiscard]] static bool IsPrefix(clang::UnaryOperatorKind op);
 
   [[nodiscard]] bool IsValidReplacementOperator(
--- a/src/libdredd/include/libdredd/mutation_replace_unary_operator.h
+++ b/src/libdredd/include/libdredd/mutation_replace_unary_operator.h
@@ -18,11 +18,11 @@
 #include <sstream>
 #include <string>
 #include <unordered_set>
+#include <vector>
 
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/ASTFwd.h"
 #include "clang/AST/OperationKinds.h"
-#include "clang/AST/Type.h"
 #include "clang/Lex/Preprocessor.h"
 #include "clang/Rewrite/Core/Rewriter.h"
 #include "libdredd/mutation.h"
@@ -56,13 +56,16 @@ class MutationReplaceUnaryOperator : public Mutation {
   [[nodiscard]] bool IsValidReplacementOperator(
       clang::UnaryOperatorKind op) const;
 
+  // This returns a string corresponding to the non-mutated expression.
   std::string GetExpr(clang::ASTContext& ast_context) const;
 
   std::string GetFunctionName(clang::ASTContext& ast_context) const;
 
-  void GenerateConstantInsertion(std::stringstream& new_function,
-                                 const clang::BuiltinType* exprType,
-                                 int& mutant_offset) const;
+  // Replaces unary operators with other valid unary operators.
+  void GenerateUnaryOperatorReplacement(
+      const std::string& arg_evaluated,
+      const std::vector<clang::UnaryOperatorKind>& operators,
+      std::stringstream& new_function, int& mutant_offset) const;
 
   const clang::UnaryOperator& unary_operator_;
 };
--- a/src/libdredd/include/libdredd/mutation_replace_unary_operator.h
+++ b/src/libdredd/include/libdredd/mutation_replace_unary_operator.h
@@ -56,7 +56,7 @@ class MutationReplaceUnaryOperator : public Mutation {
   [[nodiscard]] bool IsValidReplacementOperator(
       clang::UnaryOperatorKind op) const;
 
-  std::string getExpr(clang::ASTContext& ast_context) const;
+  std::string GetExpr(clang::ASTContext& ast_context) const;
 
   std::string GetFunctionName(clang::ASTContext& ast_context) const;
 
--- a/src/libdredd/include_private/include/libdredd/mutate_visitor.h
+++ b/src/libdredd/include_private/include/libdredd/mutate_visitor.h
@@ -50,8 +50,9 @@ class MutateVisitor : public clang::RecursiveASTVisitor<MutateVisitor> {
   bool TraverseConstantArrayTypeLoc(
       clang::ConstantArrayTypeLoc constant_array_type_loc);
 
-  // Overridden to avoid mutating variable array size expressions in C++
-  // (because lambdas cannot appear in such expressions).
+  // Overridden to avoid mutating array sizes that are derived from template
+  // parameters, because after template instantiation these lead to either
+  // constant or variable-sized arrays, neither of which can be mutated in C++.
   bool TraverseVariableArrayTypeLoc(
       clang::VariableArrayTypeLoc variable_array_type_loc);
 
--- a/src/libdredd/include_private/include/libdredd/mutate_visitor.h
+++ b/src/libdredd/include_private/include/libdredd/mutate_visitor.h
@@ -16,11 +16,15 @@
 #define LIBDREDD_MUTATE_VISITOR_H
 
 #include <memory>
+#include <set>
 #include <unordered_set>
 #include <vector>
 
+#include "clang/AST/Decl.h"
 #include "clang/AST/DeclBase.h"
 #include "clang/AST/Expr.h"
+#include "clang/AST/ExprCXX.h"
+#include "clang/AST/LambdaCapture.h"
 #include "clang/AST/RecursiveASTVisitor.h"
 #include "clang/AST/Stmt.h"
 #include "clang/AST/TemplateBase.h"
@@ -61,6 +65,18 @@ class MutateVisitor : public clang::RecursiveASTVisitor<MutateVisitor> {
   bool TraverseTemplateArgumentLoc(
       clang::TemplateArgumentLoc template_argument_loc);
 
+  // Overridden to avoid mutating lambda capture expressions, because the code
+  // that can occur in a lambda capture expression is very limited and in
+  // particular cannot include other lambdas.
+  bool TraverseLambdaCapture(clang::LambdaExpr* lambda_expr,
+                             const clang::LambdaCapture* lambda_capture,
+                             clang::Expr* init);
+
+  // Overridden to avoid mutating expressions occurring as default values for
+  // parameters, because the code that can occur in default values is very
+  // limited and cannot include lambdas in general.
+  bool TraverseParmVarDecl(clang::ParmVarDecl* parm_var_decl);
+
   bool VisitUnaryOperator(clang::UnaryOperator* unary_operator);
 
   bool VisitBinaryOperator(clang::BinaryOperator* binary_operator);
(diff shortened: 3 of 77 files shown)


JamesLee-Jones:
I still need to write new test cases and update the current ones, but it would still be useful to get a review while doing that.

afd:
I haven't looked at the code yet but I tried it on some examples and it appears to be working well. A few observations:

(1)

At first I was concerned that we don't replace `x` in:

```
int main() {
  int x;
  x;
}
```

but I suppose that's because it is an l-value - can you check please?

I suppose that in C++ mode we could insert the ++ and -- unary operators before an l-value. This could be worth doing when passing l-value arguments to functions in particular (in which case I'd suggest we do it everywhere, and open an issue for adding heuristics to avoid it in certain contexts - e.g., changing x to ++x in "x = 42" would be redundant).

(2)

In this:

```
int main() {
  bool x;
  x && x;
}
```

I see that we do not consider replacing each occurrence of `x` with "true" and "false" when we mutate them. Same thing in:

```
void foo(bool);

int main() {
  bool x;
  foo(x);
}
```

I think we should consider `true` and `false` as replacements for booleans, doing it everywhere by default and maybe later adding some heuristics to suppress it in particular contexts (e.g., it's redundant to change "x && y" to "true && y" and to also have a mutant that changes the whole thing to "y".

Would you be able to look into these things, and also add support for the `~` operator as you already mentioned?

src/libdredd/include/libdredd/mutation_replace_unary_operator.h
```diff
@@ -54,8 +56,14 @@ class MutationReplaceUnaryOperator : public Mutation {
   [[nodiscard]] bool IsValidReplacementOperator(
       clang::UnaryOperatorKind op) const;
 
+  std::string getExpr(clang::ASTContext& ast_context) const;
+
```
afd, line 60: Should be called `GetExpr` to conform to Google style.

src/libdredd/src/mutate_visitor.cc
```diff
@@ -278,6 +279,29 @@ bool MutateVisitor::VisitBinaryOperator(
...
 }
 
+bool MutateVisitor::VisitExpr(clang::Expr* expr) {
+  if (!IsInFunction()) {
+    // Only consider mutating expressions that occur inside functions.
+    return true;
+  }
+
+  // For unary and binary operators, unary operator insertion and
+  // statement replacement are handled in these classes.
```
afd, line 289: I wouldn't say "in these classes" - not clear from the context what those refer to. How about saying that unary and binary operators are intercepted separately?

src/libdredd/src/mutate_visitor.cc
```diff
@@ -278,6 +279,29 @@ bool MutateVisitor::VisitBinaryOperator(
...
+  if (llvm::dyn_cast<clang::BinaryOperator>(expr) != nullptr ||
+      llvm::dyn_cast<clang::UnaryOperator>(expr) != nullptr) {
+    return true;
+  }
+
+  if (expr->isLValue() ||
```
afd, line 295: The decision to ignore l-values deserves a comment. Also, as per my thoughts from having used this a bit, I think unary operator insertion could still be valuable for non-boolean l-values in C++ (using the ++ and -- operators).

src/libdredd/src/mutation_replace_unary_operator.cc
```diff
@@ -218,17 +285,22 @@ std::string MutationReplaceUnaryOperator::GenerateMutatorFunction(
...
+  const clang::BuiltinType* exprType =
+      unary_operator_.getType()->getAs<clang::BuiltinType>();
+  if ((exprType->isBooleanType() &&
+       unary_operator_.getOpcode() != clang::UnaryOperatorKind::UO_LNot) ||
+      (exprType->isInteger() && !exprType->isBooleanType() &&
+       !unary_operator_.isLValue())) {
+    new_function << "  if (__dredd_enabled_mutation(local_mutation_id + "
+                 << mutant_offset << ")) return !(" << getExpr(ast_context)
+                 << ");\n";
+    mutant_offset++;
```
afd, line 297: From the diff it wasn't obvious to me why this code had changed (but I'm getting a bit tired!)

src/libdredd/include/libdredd/mutation_replace_binary_operator.h
```diff
@@ -59,6 +61,12 @@ class MutationReplaceBinaryOperator : public Mutation {
   [[nodiscard]] bool IsValidReplacementOperator(
       clang::BinaryOperatorKind op) const;
 
+  std::string getExpr(clang::ASTContext& ast_context) const;
```
afd, line 64: Should be `GetExpr` to conform to the Google style.

Also I think this deserves a comment. Is it meant to return a string corresponding to the non-mutated expression (because that crops up a few times)?

afd (changes requested):
Looks very good - a few comments for now. I'll do a more careful review in the next round.

src/libdredd/include/libdredd/mutation_replace_unary_operator.h
JamesLee-Jones: Done.

src/libdredd/src/mutate_visitor.cc
JamesLee-Jones: Changed.

src/libdredd/src/mutate_visitor.cc
JamesLee-Jones: I have changed this to follow your comments above.

src/libdredd/src/mutation_replace_unary_operator.cc
JamesLee-Jones: I think the diff has messed up a little here. The changed code is for inserting the `!` unary operator where as the code that was removed was the previous way for returning the unmutated expression. The latter has now been changed to use GetExpr and is further down.

src/libdredd/include/libdredd/mutation_replace_binary_operator.h
JamesLee-Jones: Changed.

src/libdredd/include/libdredd/mutation_replace_binary_operator.h
```diff
@@ -59,6 +61,13 @@ class MutationReplaceBinaryOperator : public Mutation {
   [[nodiscard]] bool IsValidReplacementOperator(
       clang::BinaryOperatorKind op) const;
 
+  // This returns a string corresponding to the non-mutated expression.
+  std::string getExpr(clang::ASTContext& ast_context) const;
```
afd, line 65: The name should be `GetExpr` - see https://google.github.io/styleguide/cppguide.html#Function_Names

src/libdredd/include/libdredd/mutation_replace_unary_operator.h
```diff
@@ -54,8 +56,14 @@ class MutationReplaceUnaryOperator : public Mutation {
   [[nodiscard]] bool IsValidReplacementOperator(
       clang::UnaryOperatorKind op) const;
 
+  std::string GetExpr(clang::ASTContext& ast_context) const;
```
afd, line 59: Add a comment saying what this does, similar to what you did in MutationReplaceBinaryOperator.

src/libdredd/src/mutate_visitor.cc
```diff
@@ -278,6 +280,38 @@ bool MutateVisitor::VisitBinaryOperator(
...
+  // There is no useful way to mutate L-Value boolean expressions.
+  if (expr->isLValue() && expr->getType()->isBooleanType()) {
+    return true;
+  }
+
+  // We only want to mutate L-Values in C++ mode.
+  if (expr->isLValue() &&
+      !compiler_instance_.getASTContext().getLangOpts().CPlusPlus) {
+    return true;
+  }
```
afd, line 304: Is this true? I can't remember quite when things are l-values in C - but do you have an example of something that's an l-value in C that we specifically don't want to mutate? Is the point that no basic operators in C return l-values? If you could elaborate on the comment a little that would be good. (Or if it turns out there are legit cases for l-values being mutated in C, please revisit.)

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -269,10 +330,20 @@ std::string MutationReplaceBinaryOperator::GenerateMutatorFunction(
...
+  const clang::BuiltinType* exprType =
+      binary_operator_.getType()->getAs<clang::BuiltinType>();
+  if (exprType->isBooleanType() ||
+      (exprType->isInteger() && !exprType->isBooleanType() &&
+       !binary_operator_.isLValue())) {
+    new_function << "  if (__dredd_enabled_mutation(local_mutation_id + "
+                 << mutant_offset << ")) return !(" << getExpr(ast_context)
+                 << ");\n";
+    mutant_offset++;
+  }
```
afd, line 342: This handles inserting the `!` unary operator. Should any of the other unary operators be considered for insertion?

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -186,6 +205,48 @@ std::string MutationReplaceBinaryOperator::GetFunctionName(
   return result;
 }
 
+void MutationReplaceBinaryOperator::GenerateConstantInsertion(
+    std::stringstream& new_function, const clang::BuiltinType* exprType,
+    int& mutant_offset) const {
```
afd, line 210: It is good practice to put input-only parameters before output parameters:

https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

Here, `new_function` is an output parameter as it gets mutated (as is `mutant_offset`), so it should come after the input parameters.

Also: you sould make `exprType` a const reference rather than a const pointer if that's workable - see the style link above for some good advice on this. Typically non-optional parameters should be references (so that you can't pass null).

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -269,10 +330,20 @@ std::string MutationReplaceBinaryOperator::GenerateMutatorFunction(
...
+  if (exprType->isBooleanType() ||
+      (exprType->isInteger() && !exprType->isBooleanType() &&
+       !binary_operator_.isLValue())) {
+    new_function << "  if (__dredd_enabled_mutation(local_mutation_id + "
+                 << mutant_offset << ")) return !(" << getExpr(ast_context)
+                 << ");\n";
+    mutant_offset++;
+  }
+
+  GenerateConstantInsertion(new_function, exprType, mutant_offset);
```
afd, line 344: I feel like it might be clearer to also extract the code for operator replacement and unary operator insertion into functions, so that you'd have:

```
GenerateBinaryOperatorReplacement(...);
GenerateArgumentReplacement(...);
GenerateUnaryOperatorInsertion(...);
GenerateConstantReplacement(...);
```

The first would be all the logic for replacing one binary operator with another. The second the logic for replacing the operator with one of its arguments. The third the logic for inserting a unary operator. And the fourth would be replacing with a constant.

As mentioned below, I actually think `GenerateUnaryOperatorInsertion` and `GenerateConstantReplacement` could be static methods of `MutationReplaceExpr`, which get invoked from all three classes as I guess the logic should be the same.

BTW: in writing this comment it occurred to me that "ConstantInsertion" isn't really the right words, as it's a replacement, not an insertion.

src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -0,0 +1,221 @@
...
+    result += "Int";
+  } else if (expr_.getType()->isFloatingType()) {
+    result += "Float";
+  } else {
+    assert(false && "Unsupported opcode");
+  }
+
+  if (expr_.isLValue()) {
+    result += "_lvalue";
+  }
```
afd, line 52: Please make sure that ultimately there is a single file test case that relies on this distinction - i.e. one for which there would be a name clash if `lvalue` was not present.

src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -0,0 +1,221 @@
...
+  if (!expr_.isLValue() &&
+      (exprType->isBooleanType() || exprType->isInteger())) {
+    new_function << "  if (__dredd_enabled_mutation(local_mutation_id + "
+                 << mutant_offset << ")) return !(" << arg_evaluated << ");\n";
+    mutant_offset++;
+
+    new_function << "  if (__dredd_enabled_mutation(local_mutation_id + "
+                 << mutant_offset << ")) return ~(" << arg_evaluated << ");\n";
+    mutant_offset++;
+  }
```
afd, line 102: As per my comment on binary expression mutation, I think it would be good to extract this into a function.

And I suppose all the logic here, to decide when to insert a unary, should be the same as the logic in the binary and unary operator classes.

We could either do a refactoring in a follow-up PR, or you could make this a static method of  this class and have the other classes call it.

src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -0,0 +1,221 @@
...
+
+  // Insert constants for integer expressions
```
afd, line 104: It's "replace", not "insert". And you appear to handle more than just integer expressions.

src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -0,0 +1,221 @@
...
+      new_function << "  if (__dredd_enabled_mutation(local_mutation_id + "
+                   << mutant_offset << ")) return 1.0;\n";
+      mutant_offset++;
+
+      // Replace floating point expression with -1.0
+      new_function << "  if (__dredd_enabled_mutation(local_mutation_id + "
+                   << mutant_offset << ")) return -1.0;\n";
+      mutant_offset++;
+    }
+  }
```
afd, line 153: This could all be extracted into its own method. And again this could be a static method that the other classes call.

src/libdredd/src/mutation_replace_unary_operator.cc
```diff
@@ -39,6 +39,31 @@ bool MutationReplaceUnaryOperator::IsPrefix(clang::UnaryOperatorKind op) {
   return op != clang::UO_PostInc && op != clang::UO_PostDec;
 }
 
+std::string MutationReplaceUnaryOperator::GetExpr(
```
afd, line 42: Please apply analogous changes to this class as per my suggestions for "binary".

afd (changes requested):
Thanks for the changes. OK, I've done a more careful review. There are some minor things that need fixing, and now I've seen the whole thing I think there's room for some refactoring to make the code a bit easier to navigate.

I suggest you read all the comments in this review before you start acting on them, as my thinking got clearer the further along that I got!

src/libdredd/include/libdredd/mutation_replace_binary_operator.h
JamesLee-Jones: Done.

src/libdredd/include/libdredd/mutation_replace_unary_operator.h
JamesLee-Jones: Done.

src/libdredd/src/mutate_visitor.cc
JamesLee-Jones: I was incorrect so have changed it to mutate l-values in C.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: During refactoring I have added the `~` operator to also be inserted.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: I have done the refactoring described above with `GenerateUnaryOperatorInsertion` and `GenerateConstantReplacement` as static methods in `MutationReplaceExpr`.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_expr.cc
JamesLee-Jones: Done as part of the refactoring described above.

src/libdredd/src/mutation_replace_expr.cc
JamesLee-Jones: Amended.

src/libdredd/src/mutation_replace_unary_operator.cc
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_expr.cc
JamesLee-Jones: Done as part of the refactoring described above.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -232,6 +267,32 @@ std::string MutationReplaceBinaryOperator::GenerateMutatorFunction(
...
+  new_function << "  return " << GetExpr(ast_context) << ";\n";
+  new_function << "}\n\n";
+
+  // The function captures |mutant_offset| different mutations, so bump up
+  // the mutation id accordingly.
+  mutation_id += mutant_offset;
+
+  return new_function.str();
+}
+void MutationReplaceBinaryOperator::GenerateBinaryOperatorReplacement(
```
afd, line 292: Leave a space between the closing brace of the previous function and this opening brace.

Nit: I suggest placing this function right after GenerateArgumentReplacement, since they are analogous.

src/libdredd/src/mutate_visitor.cc
```diff
@@ -278,6 +279,32 @@ bool MutateVisitor::VisitBinaryOperator(
   return true;
 }
 
+bool MutateVisitor::VisitExpr(clang::Expr* expr) {
+  if (!IsInFunction()) {
+    // Only consider mutating expressions that occur inside functions.
+    return true;
+  }
+
```
afd, line 287: We need a check that the expression starts and ends in the main source file - see analogous code in the visit methods for unary and binary.

This test case currently leads to a crash:

```
#define E (1, 2, 3)

void foo(int, int, int);

int main() {
  foo E;
}
```

because from Clang's point of view, macros count as not being in the main source file. Perhaps you could add it as a single file test case - I believe the entire `foo E` should get mutated, but the arguments `(1, 2, 3)` should not (as they occur in the macro body).

src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -0,0 +1,238 @@
...
+    const clang::BuiltinType& exprType, std::stringstream& new_function,
+    int& mutant_offset) {
+  if (expr_.isLValue()) {
+    new_function << "  if (__dredd_enabled_mutation(local_mutation_id + "
+                 << mutant_offset << ")) return ++(" << arg_evaluated << ");\n";
+    mutant_offset++;
+
+    new_function << "  if (__dredd_enabled_mutation(local_mutation_id + "
+                 << mutant_offset << ")) return --(" << arg_evaluated << ");\n";
+    mutant_offset++;
```
afd, line 68: For booleans it seems redundant to have both of these, as we can usually assume that a boolean is either 1 or 0, in which case `--` would do a good job of flipping the boolean's value (since if it was 0, which is false, -1 will be regarded as true). Can you make it so that in the boolean case only `--` is applied, and add a comment?

afd (approved):
This looks really good! I have enclosed some final comments.

Once you've addressed them and added appropriate test cases please go ahead and merge, as long as the CI passes.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_expr.cc
JamesLee-Jones: Done.

src/libdredd/src/mutate_visitor.cc
JamesLee-Jones: Added a single file test case.

src/libdredd/src/mutation_replace_expr.cc
JamesLee-Jones: Added a single file test case.

afd:
@JamesLee-Jones looks like the CI is failing on the math example that you added - we're getting messages such as this one:

```
/home/runner/work/dredd/dredd/examples/math/math/src/exp.cc:522:355: error: binding reference of type 'int' to value of type 'const int' drops 'const' qualifier
long double Log(const int &a, const int &b) { if (!__dredd_enabled_mutation(698)) { return __dredd_replace_binary_operator_Div_long_double_long_double([&]() -> long double { return static_cast<long double>(__dredd_replace_expr_long_double([&]() -> long double { return static_cast<long double>(Log2(__dredd_replace_expr_int_lvalue([&]() -> int& { return static_cast<int&>(a); }, 680))); }, 682)); } , [&]() -> long double { return static_cast<long double>(__dredd_replace_expr_long_double([&]() -> long double { return static_cast<long double>(Log2(__dredd_replace_expr_int_lvalue([&]() -> int& { return static_cast<int&>(b); }, 685))); }, 687)); }, 690); } }
```

Would you be able to investigate what Dredd is doing to `exp.cc`? And please add a small single-file test case in response to whatever the root cause issue turns out to be.

Thanks!

src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -0,0 +1,289 @@
...
+void MutationReplaceExpr::ApplyCTypeModifiers(const clang::Expr* expr,
+                                              std::string& type) {
+  if (expr->isLValue()) {
+    type += "*";
+    clang::QualType qualified_type = expr->getType();
+    if (qualified_type.isVolatileQualified()) {
+      type = "volatile " + type;
+    }
+  }
+}
```
afd, line 216: It looks like these are identical to the same functions in replace unary (though I didn't check for sure).

If they are, then make them public static methods of MutationReplaceExpr, and have MutationReplaceUnaryExpr call them via MutationReplaceExpr::...

afd (approved):
I have suggested some changes, and there is the problem of why the CI fails on your math library - but as my comments are minor, and assuming that the CI issue turns out to be something small, I'm happy for you to merge once you've addressed my comments and sorted out the CI problems.

But if you need me to do another review if things turn out to be less straightforward just let me know.

afd:
@JamesLee-Jones the CI is failing due to Dredd mutating variables that occur in "auto", for example if we have:

```
int foo(int x) {
  if (auto y = x) {
    return y;
  }
  return 2;
}
```

then the `y` in `auto y` gets mutated, so we end with `auto __dredd_replace_expr(...)`, which isn't allowed.

I'll look into how we can work around this.

src/libdredd/include_private/include/libdredd/mutate_visitor.h
```diff
@@ -51,6 +51,11 @@ class MutateVisitor : public clang::RecursiveASTVisitor<MutateVisitor> {
   bool TraverseVariableArrayTypeLoc(
       clang::VariableArrayTypeLoc variable_array_type_loc);
 
+  // Overridden to avoid mutating variable array size expressions in C++
+  // (because lambdas cannot appear in such expressions).
```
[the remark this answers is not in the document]
JamesLee-Jones: Done.

src/libdredd/src/mutate_visitor.cc
```diff
@@ -162,6 +163,19 @@ bool MutateVisitor::TraverseVariableArrayTypeLoc(
...
+bool MutateVisitor::TraverseDependentSizedArrayTypeLoc(
+    clang::DependentSizedArrayTypeLoc dependent_sized_array_type_loc) {
+  // Prevent compilers complaining that this method could be made static, and
+  // that it ignores its parameter.
+  (void)this;
+  (void)dependent_sized_array_type_loc;
+  // Changing a value-dependent array to a non-constant-sized array is
+  // problematic in C, and in C++ lambdas cannot be used in array size
+  // expressions. For simplicity, don't try to mutate value-dependent array
+  // sizes.
```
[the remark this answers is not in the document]
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_expr.cc
JamesLee-Jones: Refactored.

src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -0,0 +1,314 @@
...
+
+void MutationReplaceExpr::GenerateUnaryOperatorInsertion(
+    const std::string& arg_evaluated, const clang::Expr& expr_,
+    const clang::BuiltinType& exprType, std::stringstream& new_function,
+    int& mutant_offset) {
+  if (expr_.isLValue() && !(expr_.getType().isConstQualified() ||
+                            expr_.getType()->isBooleanType())) {
+    // In the boolean case, ++ is redundant since -- is sufficient for flipping
+    // the booleans value.
```
afd, line 72: @JamesLee-Jones please check whether this is needed and get rid of it if it's not.

src/libdredd/include_private/include/libdredd/mutate_visitor.h
```diff
@@ -46,23 +50,47 @@ class MutateVisitor : public clang::RecursiveASTVisitor<MutateVisitor> {
   bool TraverseConstantArrayTypeLoc(
       clang::ConstantArrayTypeLoc constant_array_type_loc);
 
-  // Overridden to avoid mutating variable array size expressions in C++
-  // (because lambdas cannot appear in such expressions).
+  // Overridden to avoid mutating array sizes that are derived from template
+  // parameters, because after template instantiation these lead to either
+  // constant or variable-sized arrays, neither of which can be mutated in C++.
```
afd, line 55: @JamesLee-Jones looks like the comment for this method and the next need to be swapped.

afd (approved):
@JamesLee-Jones please make the final changes and then we can merge!

src/libdredd/src/mutation_replace_expr.cc
JamesLee-Jones: Done.

PR #87, opened 2022-09-01
Avoid problems with nested replacements

afd:
This change uses the rewriter's insertion and deletion functions
instead of it's functions for text replacement, to overcome problems
with nested replacement.

### DIFF
--- a/src/libdredd/CMakeLists.txt
+++ b/src/libdredd/CMakeLists.txt
@@ -19,8 +19,8 @@ add_library(
   include/libdredd/mutation.h
   include/libdredd/mutation_remove_statement.h
   include/libdredd/mutation_replace_binary_operator.h
-  include/libdredd/new_mutate_frontend_action_factory.h
   include/libdredd/mutation_replace_unary_operator.h
+  include/libdredd/new_mutate_frontend_action_factory.h
   include_private/include/libdredd/mutate_ast_consumer.h
   include_private/include/libdredd/mutate_visitor.h
   include_private/include/libdredd/util.h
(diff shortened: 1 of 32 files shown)


JamesLee-Jones (approved):
Looks good to me and is still fairly intuitive. Feel free to merge provided the tests pass.

afd:
I haven't looked at the code yet but I tried it on some examples and it appears to be working well. A few observations:

(1)

At first I was concerned that we don't replace `x` in:

```
int main() {
  int x;
  x;
}
```

but I suppose that's because it is an l-value - can you check please?

I suppose that in C++ mode we could insert the ++ and -- unary operators before an l-value. This could be worth doing when passing l-value arguments to functions in particular (in which case I'd suggest we do it everywhere, and open an issue for adding heuristics to avoid it in certain contexts - e.g., changing x to ++x in "x = 42" would be redundant).

(2)

In this:

```
int main() {
  bool x;
  x && x;
}
```

I see that we do not consider replacing each occurrence of `x` with "true" and "false" when we mutate them. Same thing in:

```
void foo(bool);

int main() {
  bool x;
  foo(x);
}
```

I think we should consider `true` and `false` as replacements for booleans, doing it everywhere by default and maybe later adding some heuristics to suppress it in particular contexts (e.g., it's redundant to change "x && y" to "true && y" and to also have a mutant that changes the whole thing to "y".

Would you be able to look into these things, and also add support for the `~` operator as you already mentioned?

afd:
Oops - did I merge this by mistake?!!

PR #98, opened 2022-09-08
Avoid redundant binary operator mutations

### LINKED ISSUE #89: Exploit constant folding to avoid redundant binary operator mutations
Clang allows us to see whether an expression can be evaluated as an int, float or bool, using code such as:

```
  clang::Expr::EvalResult eval_result;
  if (binary_operator_.EvaluateAsInt(eval_result, ast_context)) {
     ...
  }
```

We can avoid various binary operator mutations by exploiting such knowledge. For the int case, for example

In the case where the expression is `x op 0`, we can omit:
* (_, +): covered by LHS (i.e., when `op` is any operator (indicated by `_`) we can omit the `+` case, as we get something equivalent by replacing the expression with `x`)
* (_, -): covered by LHS
* (_, <<): covered by LHS
* (_, >>): covered by LHS
* (_, *): covered by replacing whole expression by 0
* (_, %): it leads to UB, which is also covered by /
* RHS: covered by replacing expression with 0

`0 op x`, omit:
* (_, +): covered by RHS
* (_, *): covered by 0
* (_, /): covered by 0
* (_, %): covered by 0
* (_, <<): covered by 0 (modulo UB)
* (_, >>): covered by 0 (modulo UB)
* LHS: covered by replacing expression with 0

`x op 1`, omit:
* (_, *): covered by LHS
* (_, /): covered by RHS
* RHS: covered by replacing expression with 1
Note: keep (_, %) - even though it will be 0 or 1, this may change each time the program point is reached so it is not covered by constant replacement

`1 op x`, omit:
* (_, *): covered by RHS
* LHS: covered by replacing expression with 1

`x op -1`, omit:
* RHS: covered by replacing whole expression with -1

`-1 op x`, omit:
* LHS: covered by replacing whole expression with -1

Similar cases apply for floats, though we might want to take care about +0.0 and -0.0 being different values.

There are also cases for booleans that should be considered.

@JamesLee-Jones for info.

JamesLee-Jones:
Exploit constant folding to reduce the number of mutations applied to binary operators in the case when one of the operands is 0, 1 or -1.

Fixes #89.


afd (commented):
Thanks for getting started on this! Some comments after a quick look - will look properly tomorrow.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -59,6 +61,47 @@ std::string MutationReplaceBinaryOperator::GetExpr(
...
+      binary_operator_.getRHS()->EvaluateAsInt(eval_result, ast_context);
+
+  // In the following cases, the replacement is equivalent to either replacement
+  // with a constant or argument replacement.
+  // See https://github.com/mc-imperial/dredd/issues/89 for a more in-depth
+  // explanation.
+  if (rhs_is_int && llvm::APSInt::isSameValue(eval_result.Val.getInt(),
+                                              llvm::APSInt::get(0))) {
+    if (op == clang::BO_Add || op == clang::BO_Sub || op == clang::BO_Shl ||
+        op == clang::BO_Shr || op == clang::BO_Mul || op == clang::BO_Rem) {
```
afd, line 77: I am not sure about the "rem" case - remainder by zero is an undefined behaviour so it's not covered by other cases.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -59,6 +61,47 @@ std::string MutationReplaceBinaryOperator::GetExpr(
...
+bool MutationReplaceBinaryOperator::IsRedundantReplacementOperator(
+    clang::BinaryOperatorKind op, clang::ASTContext& ast_context) const {
+  clang::Expr::EvalResult eval_result;
+  bool rhs_is_int =
+      binary_operator_.getRHS()->EvaluateAsInt(eval_result, ast_context);
+
+  // In the following cases, the replacement is equivalent to either replacement
+  // with a constant or argument replacement.
+  // See https://github.com/mc-imperial/dredd/issues/89 for a more in-depth
+  // explanation.
```
afd, line 73: I would get rid of this reference to the issue. But I think it would be worth adding brief comments to note what each case is covered by.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -203,31 +246,88 @@ std::string MutationReplaceBinaryOperator::GetFunctionName(
...
 void MutationReplaceBinaryOperator::GenerateArgumentReplacement(
     const std::string& arg1_evaluated, const std::string& arg2_evaluated,
-    std::stringstream& new_function, int& mutant_offset) const {
+    clang::ASTContext& ast_context, std::stringstream& new_function,
+    int& mutant_offset) const {
   if (!binary_operator_.isAssignmentOp()) {
+    clang::Expr::EvalResult lhs_eval_result;
+    clang::Expr::EvalResult rhs_eval_result;
+    bool rhs_is_int =
+        binary_operator_.getRHS()->EvaluateAsInt(rhs_eval_result, ast_context);
```
afd, line 289: I suggest moving the declaration of rhs_eval_result, and the call to EvaluateAsInt for RHS, until right before the block of code where you use it.

In general it's good practice to put declarations close to their use, to prevent the possibility of accidentally referring to the wrong variable.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -203,31 +246,88 @@ std::string MutationReplaceBinaryOperator::GetFunctionName(
...
+    bool lhs_is_int =
+        binary_operator_.getLHS()->EvaluateAsInt(lhs_eval_result, ast_context);
+
     // LHS
-    new_function << "  if (__dredd_enabled_mutation(local_mutation_id + "
-                 << mutant_offset << ")) return " << arg1_evaluated << ";\n";
-    mutant_offset++;
+    // These cases are equivalent to constant replacement with the respective
+    // constants
```
afd, line 295: I'd move the declarations of LHS stuff to after this comment.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -203,31 +246,88 @@ std::string MutationReplaceBinaryOperator::GetFunctionName(
...
+                   << mutant_offset << ")) return " << arg1_evaluated << ";\n";
+      mutant_offset++;
+    }
 
     // RHS
-    new_function << "  if (__dredd_enabled_mutation(local_mutation_id + "
-                 << mutant_offset << ")) return " << arg2_evaluated << ";\n";
-    mutant_offset++;
+    // These cases are equivalent to constant replacement with the respective
+    // constants
```
afd, line 309: And I'd move the declaration of RHS stuff to after this comment.

afd (changes requested):
This looks great!

Can you please look at adding a boolean command line option to dredd for disabling optimisations? I think `--no-mutation-opts` would be a good name for the flag. This should be used to set an `optimise_mutations_` field in `MutateAstConsumer`, and you should then give the `Apply` method in the `Mutation` interface a new input parameter, `optimise_mutations`, which can be queried to determine whether to optimise.

This will require quite a bit of threading through, but I think for the research it will be interesting to know the impact of these optimisations.

Here are some notes on LLVM command line arguments:

https://llvm.org/docs/CommandLine.html#boolean-arguments
https://llvm.org/docs/CommandLine.html#grouping-options-into-categories

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Didn't you note that the undefined behavior case was covered by `/`?

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -203,31 +246,88 @@ std::string MutationReplaceBinaryOperator::GetFunctionName(
                                   ->getName(ast_context.getPrintingPolicy())
                                   .str());
 
+  if (!binary_operator_.isAssignmentOp()) {
```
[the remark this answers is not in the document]
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Done.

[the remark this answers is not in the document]
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_binary_operator.cc
afd: Ah yes - you're right. When you add comments it would be good to note that.

src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -311,6 +312,7 @@ void MutationReplaceExpr::Apply(
   assert(!result && "Rewrite failed.\n");
   (void)result;
 
+  (void)optimise_mutations;
```
afd, line 315: Similar "Unused" comment and position.

src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -311,6 +312,7 @@ void MutationReplaceExpr::Apply(
   assert(!result && "Rewrite failed.\n");
   (void)result;
```
afd, line 313: Please sneak in a "keep release-mode compilers happy" comment similar to other `Apply` methods.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -58,6 +60,70 @@ std::string MutationReplaceBinaryOperator::GetExpr(
...
+    }
+  }
+
+  // In the following cases, the replacement is equivalent to either replacement
+  // with a constant or argument replacement.
+  if (rhs_is_int && llvm::APSInt::isSameValue(rhs_eval_result.Val.getInt(),
+                                              llvm::APSInt::get(0))) {
+    // When the right operand is 0: +, -, << and >> are all equivalent to
+    // replacement with the right operand; * is equivalent to replacement with
+    // the constant 0 and % is equivalent to replacement with /.
```
afd, line 92: For the `%` case, I suggest `is equivalent to replacement with / in that both cases lead to undefined behaviour.`

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -58,6 +60,70 @@ std::string MutationReplaceBinaryOperator::GetExpr(
...
+    if (op == clang::BO_Add || op == clang::BO_Sub || op == clang::BO_Shl ||
+        op == clang::BO_Shr || op == clang::BO_Mul || op == clang::BO_Rem) {
+      return true;
+    }
+  } else if (rhs_is_int &&
+             llvm::APSInt::isSameValue(rhs_eval_result.Val.getInt(),
+                                       llvm::APSInt::get(1))) {
```
afd, line 99: I would be tempted to change this from `else if` to just a fresh `if`, because the rhs evaluating to 0 and 1 are already mutually exclusive.

I would also be tempted to factor out the `rhs_is_int` condition of both cases, giving:

```
if (rhs_is_int) {
  if (the zero case) {

  }
  if (the one case) {

  }
}
```

... but this is getting very nit picky and is really a matter of taste.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -58,6 +60,70 @@ std::string MutationReplaceBinaryOperator::GetExpr(
...
+    // When the right operand is 1: * and / are equivalent to replacement by the
+    // left operand.
+    if (op == clang::BO_Mul || op == clang::BO_Div) {
+      return true;
+    }
+  }
+
+  if (lhs_is_int && llvm::APSInt::isSameValue(lhs_eval_result.Val.getInt(),
```
afd, line 107: Same comment as my nit-picky one above is relevant here.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -202,31 +268,94 @@ std::string MutationReplaceBinaryOperator::GetFunctionName(
...
+    clang::Expr::EvalResult lhs_eval_result;
+    bool lhs_is_int =
+        binary_operator_.getLHS()->EvaluateAsInt(lhs_eval_result, ast_context);
+    if (!optimise_mutations ||
+        !(lhs_is_int && (llvm::APSInt::isSameValue(lhs_eval_result.Val.getInt(),
+                                                   llvm::APSInt::get(0)) ||
+                         llvm::APSInt::isSameValue(lhs_eval_result.Val.getInt(),
+                                                   llvm::APSInt::get(1)) ||
+                         llvm::APSInt::isSameValue(lhs_eval_result.Val.getInt(),
+                                                   llvm::APSInt::get(-1))))) {
```
afd, line 324: I am worried that we might lose something when the LHS is *unsigned*, but has a constant value that would evaluate to  -1 if interpreted as signed. In that case we wouldn't do the -1 replacement in constant replacement mode. Can you dig into this a bit?

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -202,31 +268,94 @@ std::string MutationReplaceBinaryOperator::GetFunctionName(
...
+    bool rhs_is_int =
+        binary_operator_.getRHS()->EvaluateAsInt(rhs_eval_result, ast_context);
+    if (!optimise_mutations ||
+        !(rhs_is_int && (llvm::APSInt::isSameValue(rhs_eval_result.Val.getInt(),
+                                                   llvm::APSInt::get(0)) ||
+                         llvm::APSInt::isSameValue(rhs_eval_result.Val.getInt(),
+                                                   llvm::APSInt::get(1)) ||
+                         llvm::APSInt::isSameValue(rhs_eval_result.Val.getInt(),
+                                                   llvm::APSInt::get(-1))))) {
+      new_function << "  if (__dredd_enabled_mutation(local_mutation_id + "
```
afd, line 343: Similar.

afd (changes requested):
This is looking excellent! My comments are mostly nit-picky but a slightly larger one about -1 and unsigned (which I'm not sure is a real issue but would be good to check).

I would also like the changes in the tests to be handled differently so that we still test the unoptimised code paths - will have at think and get back to you on that.

src/dredd/src/main.cc
```diff
@@ -41,6 +41,10 @@ static llvm::cl::extrahelp common_help(
     clang::tooling::CommonOptionsParser::HelpMessage);
 // NOLINTNEXTLINE
 static llvm::cl::OptionCategory mutate_category("mutate options");
+// NOLINTNEXTLINE
+static llvm::cl::opt<bool> NoMutationOpts(
```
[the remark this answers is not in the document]
JamesLee-Jones: Done.

src/libdredd/include_private/include/libdredd/mutate_ast_consumer.h
```diff
@@ -48,6 +49,8 @@ class MutateAstConsumer : public clang::ASTConsumer {
   // Counter used to give each mutation a unique id; shared among AST consumers
   // for different translation units.
   int& mutation_id_;
+  // Used to disable dredd's mutations.
+  bool optimise_mutations_;
```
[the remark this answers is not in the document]
JamesLee-Jones: Done.

src/libdredd/include/libdredd/mutation.h
```diff
@@ -49,7 +49,7 @@ class Mutation {
   // avoiding redundant repeat declarations.
   virtual void Apply(
       clang::ASTContext& ast_context, const clang::Preprocessor& preprocessor,
-      int first_mutation_id_in_file, int& mutation_id,
```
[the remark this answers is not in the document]
JamesLee-Jones: Refactored.

src/libdredd/src/mutation_remove_statement.cc
```diff
@@ -95,6 +96,7 @@ void MutationRemoveStatement::Apply(
           "}");
   assert(!result && "Rewrite failed.\n");
   (void)result;  // Keep release-mode compilers happy.
+  (void)optimise_mutations;
```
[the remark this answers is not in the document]
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_unary_operator.cc
```diff
@@ -327,6 +328,7 @@ void MutationReplaceUnaryOperator::Apply(
   assert(!result && "Rewrite failed.\n");
   (void)result;  // Keep release-mode compilers happy.
 
+  (void)optimise_mutations;
```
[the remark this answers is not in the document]
JamesLee-Jones: Done.

src/libdredd/src/new_mutate_frontend_action_factory.cc
```diff
@@ -53,37 +56,42 @@ class MutateFrontendAction : public clang::ASTFrontendAction {
...
     std::unique_ptr<clang::FrontendAction> create() override {
-      return std::make_unique<MutateFrontendAction>(mutation_id_,
-                                                    processed_files_);
+      return std::make_unique<MutateFrontendAction>(
+          optimise_mutations_, mutation_id_, processed_files_);
     }
 
    private:
     int& mutation_id_;
+    bool optimise_mutations_;
```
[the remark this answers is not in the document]
JamesLee-Jones: Done.

src/libdredd/src/new_mutate_frontend_action_factory.cc
```diff
@@ -53,37 +56,42 @@ class MutateFrontendAction : public clang::ASTFrontendAction {
...
 std::unique_ptr<clang::tooling::FrontendActionFactory>
-NewMutateFrontendActionFactory(int& mutation_id) {
+NewMutateFrontendActionFactory(bool optimise_mutations, int& mutation_id) {
   class MutateFrontendActionFactory
       : public clang::tooling::FrontendActionFactory {
    public:
-    explicit MutateFrontendActionFactory(int& mutation_id)
-        : mutation_id_(mutation_id) {}
+    explicit MutateFrontendActionFactory(bool optimise_mutations,
+                                         int& mutation_id)
```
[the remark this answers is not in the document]
JamesLee-Jones: Done.

src/libdredd/src/new_mutate_frontend_action_factory.cc
```diff
@@ -53,37 +56,42 @@ class MutateFrontendAction : public clang::ASTFrontendAction {
 
  private:
   int& mutation_id_;
+  bool optimise_mutations_;
```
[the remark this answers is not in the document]
JamesLee-Jones: Done.

src/libdredd/src/new_mutate_frontend_action_factory.cc
```diff
@@ -31,8 +31,11 @@ namespace dredd {
 
 class MutateFrontendAction : public clang::ASTFrontendAction {
  public:
-  MutateFrontendAction(int& mutation_id, std::set<std::string>& processed_files)
-      : mutation_id_(mutation_id), processed_files_(processed_files) {}
+  MutateFrontendAction(bool optimise_mutations, int& mutation_id,
```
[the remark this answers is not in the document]
JamesLee-Jones: Done.

src/libdredd/src/new_mutate_frontend_action_factory.cc
```diff
@@ -53,37 +56,42 @@ class MutateFrontendAction : public clang::ASTFrontendAction {
...
   };
 
-  return std::make_unique<MutateFrontendActionFactory>(mutation_id);
+  return std::make_unique<MutateFrontendActionFactory>(optimise_mutations,
+                                                       mutation_id);
 }
 
 std::unique_ptr<clang::ASTConsumer> MutateFrontendAction::CreateASTConsumer(
     clang::CompilerInstance& ci, llvm::StringRef file) {
   (void)file;
```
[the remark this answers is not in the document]
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_expr.cc
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Done.

afd:
@JamesLee-Jones Here's what I would like you to do for the single file tests:

* Change the check and regenerate scripts (both the "all" and "one" versions) so that they do the "run dredd then compile" step twice: once with optimizations enabled (no flag), and once with no optimizations (the flag).
* In the case where optimizations are enabled they should check against the `blah.c/cc.expected` file (or generate this if it doesn't exist, in the `regenerate` cases)
* In the case where optimizations are disabled they should check against a `blah.c/cc.noopt.expected` (or generate this if it doesn't exist)

This will lead to many cases where `blah.c/cc.expected` is identical to `blah.c/cc.noopt.expected`, but I think that's OK because:

* as we implement more optimisations the differences will decrease
* I don't think it matters if there's some redundancy (I toyed with the idea of having the `.noopt` file only be expected / generated if it's different from the other expectation file, but that makes the testing code more complex)

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: I agree that this is probably a nicer way of writing it so I've refactored it.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Same reply as above.

src/libdredd/src/new_mutate_frontend_action_factory.cc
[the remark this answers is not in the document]
JamesLee-Jones: Done.

src/libdredd/include_private/include/libdredd/mutate_visitor.h
```diff
@@ -38,7 +38,8 @@ namespace dredd {
 
 class MutateVisitor : public clang::RecursiveASTVisitor<MutateVisitor> {
  public:
-  explicit MutateVisitor(const clang::CompilerInstance& compiler_instance);
+  explicit MutateVisitor(const clang::CompilerInstance& compiler_instance,
+                         bool optimise_mutations);
```
afd, line 42: Remove `explicit` as there are multiple arguments.

src/libdredd/include/libdredd/mutation_replace_expr.h
```diff
@@ -32,14 +32,18 @@ class MutationReplaceExpr : public Mutation {
...
+      bool optimise_mutations, int first_mutation_id_in_file, int& mutation_id,
       clang::Rewriter& rewriter,
       std::unordered_set<std::string>& dredd_declarations) const override;
 
   static void ApplyCppTypeModifiers(const clang::Expr* expr, std::string& type);
 
   static void ApplyCTypeModifiers(const clang::Expr* expr, std::string& type);
 
+  // Check if an expression is equivalent to a constant.
+  static bool ExprIsEquivalentTo(clang::Expr* expr, int x,
```
afd, line 44: Could `expr` be changed to `const clang::Expr& expr`?

Please change `x` to have a more descriptive name (both in the .h and .cc file).

afd (changes requested):
Looking great! A few small things.

src/libdredd/include_private/include/libdredd/mutate_visitor.h
JamesLee-Jones: Done.

src/libdredd/include/libdredd/mutation_replace_expr.h
JamesLee-Jones: Changed both.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: After experimenting, this fortunately doesn't appear to be an issue. Even when an unsigned integer operand could be evaluated as -1, we still replace the expression with both operands. I have added a test case to check this.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Same as above.

afd (approved):
Nice work!

PR #100, opened 2022-09-09
Do not mutate parentheses expressions

### LINKED ISSUE #99: Do not mutate parentheses expressions
There is no point mutating an expression `(e)` since we will mutate `e`.

afd:
Avoids mutating expressions of the form (e), since mutating e suffices.

Fixes #99.


JamesLee-Jones (commented):
Looks good! Feel free to merge if the tests pass.

JamesLee-Jones (approved):

PR #107, opened 2022-09-12
Support logical operators in C

afd:
Provides support for C logical operators via an alternative rewriting.


JamesLee-Jones (approved):
Looks good! LGTM

PR #110, opened 2022-09-16
Avoid redundant unary operator mutations

### LINKED ISSUE #90: Exploit constant folding to avoid redundant unary operator mutations
Like #89, but for unary operators.

JamesLee-Jones:
Exploit constant folding to reduce the number of mutations applied to unary operators in the case when the operand is 0, 1 or -1.

Fixes #90.


src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -63,9 +63,6 @@ std::string MutationReplaceBinaryOperator::GetExpr(
 
 bool MutationReplaceBinaryOperator::IsRedundantReplacementOperator(
     clang::BinaryOperatorKind op, clang::ASTContext& ast_context) const {
-  clang::Expr::EvalResult rhs_eval_result;
-  clang::Expr::EvalResult lhs_eval_result;
-
```
afd, line 68: I am curious as to why the CI passed last time - I would have expected a complaint about these not being used. Good spot this round.

src/libdredd/src/mutation_replace_unary_operator.cc
```diff
@@ -163,13 +163,21 @@ std::string MutationReplaceUnaryOperator::GetFunctionName(
...
                                   .str());
 
+  // Since the optimised cases are the same for both op 0 and op 1, we only need
+  // to distinguish in the case op -1
+  if (optimise_mutations &&
+      MutationReplaceExpr::ExprIsEquivalentTo(*unary_operator_.getSubExpr(), -1,
+                                              ast_context)) {
+    result += "_minus_one";
+  }
+
```
afd, line 173: I notice the CI is failing due to redefinitions. I didn't look into it, but perhaps this isn't quite right.

src/libdredd/src/mutation_replace_unary_operator.cc
```diff
@@ -221,12 +230,36 @@ std::string MutationReplaceUnaryOperator::GenerateMutatorFunction(
...
+  if (MutationReplaceExpr::ExprIsEquivalentTo(*unary_operator_.getSubExpr(), -1,
+                                              ast_context) &&
+      op == clang::UO_Minus) {
+    return true;
+  }
+
+  return false;
+}
```
afd, line 252: Can you add some comments to justify these cases? I can see tha -0 is equivalent, and -1 is covered by replacing with -1 in the signed case (do we apply - to unsigned?)

In general some comments would be good.

src/libdredd/src/mutation_replace_unary_operator.cc
```diff
@@ -240,9 +273,18 @@ void MutationReplaceUnaryOperator::GenerateUnaryOperatorReplacement(
...
+                                              ast_context) ||
+      MutationReplaceExpr::ExprIsEquivalentTo(*unary_operator_.getSubExpr(), 1,
+                                              ast_context) ||
+      MutationReplaceExpr::ExprIsEquivalentTo(*unary_operator_.getSubExpr(), -1,
+                                              ast_context)) {
+    new_function << "  if (__dredd_enabled_mutation(local_mutation_id + "
+                 << mutant_offset << ")) return " + arg_evaluated + ";\n";
+    mutant_offset++;
+  }
 }
```
afd, line 288: A comment here to justify what these are covered by would be good.

afd (approved):
This looks very good. I have some comments, mainly about commentary :)

Happy for you to merge once you've addressed those and worked out why the CI is failing.

src/libdredd/src/mutation_replace_binary_operator.cc
afd: (This one doesn't need action so resolve it once digested. I'm just a bit mystified by why the compiler sometimes complains about unused variables but not always.)

JamesLee-Jones: It is very confusing!

src/libdredd/src/mutation_replace_unary_operator.cc
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_unary_operator.cc
JamesLee-Jones: Added comments.

JamesLee-Jones: @afd I was concerned about that being the case, I think i noticed in one of the test cases that in an expression like `unsigned int x = ~5`, the expression may be replaced with a negative number. I'll take a look and check.

src/libdredd/src/mutation_replace_unary_operator.cc
JamesLee-Jones: It seems that this may have been the issue. I've changed it to add the specifier in the three unique cases.

src/libdredd/src/mutation_replace_unary_operator.cc
JamesLee-Jones: I have filled this as an issue #115.

PR #111, opened 2022-09-17
Represent mutations as a tree, to allow easy detection of mutually-exclusive mutants when using Dredd to find interesting test cases

afd:
With this change, Dredd now emits, in JSON format, information about the mutations that are applied in tree form. Mutations at the same tree node, or with a parent-child relationship, are deemed mutually exclusive. A tool that uses Dredd can therefore enable sets of mutually-independent mutations.


afd:
@JamesLee-Jones unfortunately I had to change slightly the order in which AST nodes get visited for this to work, which leads to lots of changes in the single file tests - these should all just be changes to the specific ids that are used for mutations.

JamesLee-Jones (approved):
Looks good, nice change! LGTM

PR #116, opened 2022-09-18
Refactor scripts for handling single-file tests

### LINKED ISSUE #113: Refactor scripts for handling single-file test cases
There is now quite a lot of duplication between these scripts; some of it should be factored out. In particular the "single file" scripts should be invoked by the multi-file scripts (but the copying of the Dredd binary should not be performed on every single test).

afd:
Avoids duplication in the scripts for checking and regenerating
single-file tests and expectations.
    
Fixes #113.

### DIFF
--- a/scripts/check_one_single_file_test.sh
+++ b/scripts/check_one_single_file_test.sh
@@ -28,13 +28,15 @@ if [ -z "${DREDD_SKIP_COPY_EXECUTABLE+x}" ]
 then
   # Ensure that Dredd is in its installed location. This depends on a
   # debug build being available
-  cp build-Debug/src/dredd/dredd ${DREDD_INSTALLED_EXECUTABLE}
+  cp build-Debug/src/dredd/dredd "${DREDD_INSTALLED_EXECUTABLE}"
 fi
 
 f="${DREDD_REPO_ROOT}/${1}"
 
 for opt in "opt" "noopt"
 do
+  # Determine whether optimisations should be disabled or not, and whether the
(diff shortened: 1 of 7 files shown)


afd:
@JamesLee-Jones I am not very good with bash, so if you have more experience here and reckon there's anything here that is poor bash practice please let me know!

scripts/check_one_single_file_test.sh
```diff
@@ -18,47 +18,63 @@ set -e
...
   # that it can be mutated without affecting the original
   cp $f .
   copy_of_f=$(basename $f)
-  # Mutate the test case using Dredd without optimisations
-  ${DREDD_INSTALLED_EXECUTABLE} --no-mutation-opts --mutation-info-file temp.json ${copy_of_f} --
-  # Check that the mutated test case is as expected
-  diff ${copy_of_f} $f.noopt.expected
+
+  # Mutate the test case using Dredd with optimisations
+  ${DREDD_INSTALLED_EXECUTABLE} ${DREDD_EXTRA_ARGS} --mutation-info-file temp.json ${copy_of_f} --
```
JamesLee-Jones, line 54: This comment shouldn't be here anymore because the command will also run Dredd without optimisations.

JamesLee-Jones (approved):
These changes look good! Only one small nit about a comment but after that feel free to merge.

scripts/check_one_single_file_test.sh
afd: Fixed, thanks.

PR #120, opened 2022-09-20
Insert increment before conversion

afd:


JamesLee-Jones (approved):
Looks good to me! LGTM

PR #124, opened 2022-09-21
Restrict mutation of cast expressions

### LINKED ISSUE #121: Avoid redundant mutations before casts
Sometimes the AST will feature a nest of cast expressions. E.g.:

```
int foo() {
  unsigned x;
  return static_cast<int>(x);
}
```

will yield an explicit cast that encloses two implicit casts.

It should suffice to mutate a cast expression only when:

- its sub-expression is not a cast expression
- it is an l-value to r-value cast

afd:
Restricts mutation of cast expressions to the case where the expression inside the case is an l-value and the cast expression is an r-value. This avoids a large number of cases where literal values are implicitly case from signed to unsigned, or from double to float, and get mutated twice as a result.

Fixes #121.

### DIFF
--- a/src/libdredd/src/mutate_visitor.cc
+++ b/src/libdredd/src/mutate_visitor.cc
@@ -365,6 +365,22 @@ bool MutateVisitor::VisitExpr(clang::Expr* expr) {
     return true;
   }
 
+  if (optimise_mutations_) {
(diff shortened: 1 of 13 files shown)


JamesLee-Jones (approved):
Nice job addressing this, it looks good to me!

Issue #108, opened 2022-09-16
Consider not bitwise-negating boolean expressions


afd:
We see a lot of this warning:

```
warning: bitwise negation of a boolean expression always evaluates to 'true'; did you mean logical negation? [-Wbool-operation]
```

We should also consider other cases where mutations will end up being redundant.

afd:
I think it indeed makes sense to avoid applying `~` to a boolean expression.

@JamesLee-Jones would you be able to do a PR to fix this? I don't think this should count as an optimisation.

PR #127, opened 2022-09-21
Avoid redundant unary operator insertion

### LINKED ISSUE #92: Exploit constant folding to avoid redundant unary operator insertion
Like #89, but for the case of unary operator insertion.

For example, if `e` is a boolean that evaluates to `true`, we should not insert `!` as this is covered by replacing `e` with `false`.

JamesLee-Jones:
Exploit constant folding to avoid redundant unary operator insertion when it is equivalent to replacement with a constant. Also avoid applying `~` to boolean expressions.

Fixes #92, #108.


src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -62,23 +62,39 @@ std::string MutationReplaceExpr::GetFunctionName(
     result += "_lvalue";
   }
 
+  if (IsRedundantOperatorInsertion(expr_, ast_context)) {
+    result += "_optimised";
+  }
```
afd, line 67: This is fine if there's only one way that the generated function changes due to redundancy. If there are are multiple different ways then we will get name clashes. Can you please check? If it's fine as is, please add a reassuring comment.

src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -62,23 +62,39 @@ std::string MutationReplaceExpr::GetFunctionName(
...
                                      llvm::APSInt::get(constant));
   }
   return false;
 }
 
+bool MutationReplaceExpr::IsRedundantOperatorInsertion(
+    const clang::Expr& expr, clang::ASTContext& ast_context) {
+  bool result;
+  return (expr.getType()->isBooleanType() &&
+          expr.EvaluateAsBooleanCondition(result, ast_context)) ||
```
afd, line 88: Add a comment that we use `result` to capture that the expression evaluates to if it indeed evaluates to a constant, but that we don't actually use this value because either way it means the operator is redundant.

Separate comment: I would change it from `result` to something else, because `result` typically refers to the value that the current method will eventually return. I'd call it `unused_bool_value` or something.

afd (approved):
Looks good - a couple of small requests for comments / changes. Feel free to merge once addressed.

src/libdredd/src/mutation_replace_expr.cc
JamesLee-Jones: This does appear to be fine because this is the only optimization that depends on the expression. The other UOI optimization that is applied, omitting `~`, is dependent on the expression type being bool, so the type included in the function name should be enough to distinguish this case.

src/libdredd/src/mutation_replace_expr.cc
JamesLee-Jones: Done.

PR #128, opened 2022-09-21
Avoid redundant expression replacement

### LINKED ISSUE #91: Exploit constant folding to avoid redundant expression replacement
Like #89, but for entire expressions (e.g., if `e` evaluates to 0, we should not consider the case where `e` gets replaced by 0.

JamesLee-Jones:
Use constant folding to avoid replacing an expression with a constant when the expression already evaluates to that value. For example, if an expression e evaluates to 0, avoid replacing e with the constant 0.

Fixes #91.

### DIFF
--- a/src/libdredd/src/mutation_replace_expr.cc
+++ b/src/libdredd/src/mutation_replace_expr.cc
@@ -63,31 +63,40 @@ std::string MutationReplaceExpr::GetFunctionName(
     result += "_lvalue";
   }
 
-  // This is sufficient to prevent name clashes as this is the
-  // only unary operator insertion optimisation that depends on the
-  // input expression. ~ is omitted for all boolean expressions
-  // so the type that is included in the function name will be enough
-  // in that case.
   if (optimise_mutations) {
+    // This is sufficient to prevent name clashes as this is the
+    // only unary operator insertion optimisation that depends on the
+    // input expression. ~ is omitted for all boolean expressions
+    // so the type that is included in the function name will be enough
+    // in that case.
     if (IsRedundantOperatorInsertion(expr_, ast_context)) {
       result += "_uoi_optimised";
     }
 
-    if (ExprIsEquivalentTo(expr_, 0, ast_context)) {
-      result += "_zero";
-    } else if (ExprIsEquivalentTo(expr_, 1, ast_context)) {
-      result += "_one";
-    } else if (ExprIsEquivalentTo(expr_, -1, ast_context)) {
-      result += "_minus_one";
+    if ((expr_.getType()->isIntegerType() && !expr_.getType()->isBooleanType()) || expr_.getType()->isFloatingType()) {
+      if (ExprIsEquivalentToInt(expr_, 0, ast_context) || ExprIsEquivalentToFloat(expr_, 0, ast_context)) {
+        result += "_zero";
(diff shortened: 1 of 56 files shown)


src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -67,8 +68,18 @@ std::string MutationReplaceExpr::GetFunctionName(
   // input expression. ~ is omitted for all boolean expressions
   // so the type that is included in the function name will be enough
   // in that case.
-  if (optimise_mutations && IsRedundantOperatorInsertion(expr_, ast_context)) {
-    result += "_uoi_optimised";
+  if (optimise_mutations) {
+    if (IsRedundantOperatorInsertion(expr_, ast_context)) {
```
afd, line 72: I suggest you move the above comment so that it occurs just before (or just inside) the inner `if`, since it doesn't apply to whole block anymore.

src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -77,12 +88,25 @@ std::string MutationReplaceExpr::GetFunctionName(
...
                                      llvm::APSInt::get(constant));
   }
+
+  llvm::APFloat float_eval_result(static_cast<double>(0));
+  if (expr.getType()->isFloatingType() &&
+      expr.EvaluateAsFloat(float_eval_result, ast_context)) {
+    return float_eval_result.isExactlyValue(constant);
+  }
+
+  bool result = false;
```
afd, line 104: Rename `result` to `bool_eval_result` for consistency (even though it has a different type), and to avoid breaking the convention that `result` is used for the thing that will be returned.

src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -77,12 +88,25 @@ std::string MutationReplaceExpr::GetFunctionName(
 bool MutationReplaceExpr::ExprIsEquivalentTo(const clang::Expr& expr,
                                              int constant,
                                              clang::ASTContext& ast_context) {
```
afd, line 90: I am not sure I like the use of `int` for all three data types here.

I think it would be better to have three separate methods, `ExprIsEquivalentToInt`, `ExprIsEquivalentToFloat` and `ExprIsEquivalentToBool`, taking `int`, `double` and `bool` arguments respectively.

afd (approved):
This looks good to go, except that I would be more comfortable with three separate `IsEquivalentTo` functions; one for int, one for float/double and one for bool.

It seems hacky to me to use `1` and `0` for `true` and `false`, and if in the future we wanted to use special floating-point values, or even non-integer floating-point values, the float case would fall down.

src/libdredd/src/mutation_replace_expr.cc
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_expr.cc
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_expr.cc
JamesLee-Jones: Refactored.

PR #129, opened 2022-09-21
Do not mutate top-level expressions in blocks

### LINKED ISSUE #93: Avoid replacing a whole statement with a constant
Suppose we have a statement:

`foo();`

Replacing `foo()` a constant, `0` say, has the same effect as removing the statement, so we do not want to have both mutants.

afd:
Avoid applying general expression mutation to top-level expression statements that appear in compound statements, as there is little-to-no value in doing so.

Fixes #93.

### DIFF
--- a/src/libdredd/src/mutate_visitor.cc
+++ b/src/libdredd/src/mutate_visitor.cc
@@ -18,6 +18,8 @@
 #include <cstddef>
 #include <memory>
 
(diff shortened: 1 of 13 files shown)


JamesLee-Jones (approved):
Looks good! Feel free to merge provided the CI passes.

PR #133, opened 2022-09-21
Make unary and binary optimisations exploit float constants

### LINKED ISSUE #132: Binary and unary operator replacement should take account of floating-point constants
Reviewing #128, which appears to handle the floating-point case, made me realise that we should take account of floating point constants when considering unary and binary expressions.

JamesLee-Jones:
Exploit the float constants 0.0, 1.0 and -1.0 to simplify the set of mutations that are applied.

Fixes #132.


src/libdredd/src/mutate_visitor.cc
```diff
@@ -258,6 +258,30 @@ bool MutateVisitor::HandleUnaryOperator(clang::UnaryOperator* unary_operator) {
     return true;
   }
 
+  // There is no useful way to mutate this expression.
+  if (optimise_mutations_ && ((unary_operator->getOpcode() == clang::UO_Minus &&
```
afd, line 262: This boolean expression is too big for my taste :)

Can you please refactor to:

```
if (optimise_mutations_) {
   if (opcode is minus && (disjunction of conditions)) {

   }
   if (opcode is not && (disjunction of conditions)) {

   }
}
```

src/libdredd/src/mutate_visitor.cc
```diff
@@ -258,6 +258,30 @@ bool MutateVisitor::HandleUnaryOperator(clang::UnaryOperator* unary_operator) {
...
+                               (MutationReplaceExpr::ExprIsEquivalentToInt(
+                                    *unary_operator->getSubExpr(), 1,
+                                    compiler_instance_.getASTContext()) ||
+                                MutationReplaceExpr::ExprIsEquivalentToFloat(
+                                    *unary_operator->getSubExpr(), 1,
```
afd, line 267: For the float cases, please use a double literal - so 1.0 in this case. Please do the same for other floating-point checks.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -351,16 +352,19 @@ void MutationReplaceBinaryOperator::GenerateArgumentReplacement(
...
-      !(rhs_is_int && (llvm::APSInt::isSameValue(rhs_eval_result.Val.getInt(),
-                                                 llvm::APSInt::get(0)) ||
-                       llvm::APSInt::isSameValue(rhs_eval_result.Val.getInt(),
-                                                 llvm::APSInt::get(1)) ||
-                       llvm::APSInt::isSameValue(rhs_eval_result.Val.getInt(),
-                                                 llvm::APSInt::get(-1))))) {
+      !(MutationReplaceExpr::ExprIsEquivalentToInt(*binary_operator_.getRHS(),
+                                                   0, ast_context) ||
+        MutationReplaceExpr::ExprIsEquivalentToFloat(*binary_operator_.getRHS(),
+                                                     0, ast_context) ||
```
afd, line 359: Similarly here, and in the code below for the RHS, please use double constants where relevant.

afd (approved):
A couple of small changes requested but otherwise looks good.

src/libdredd/src/mutate_visitor.cc
JamesLee-Jones: Done.

src/libdredd/src/mutate_visitor.cc
JamesLee-Jones: Done.

src/libdredd/src/mutation_replace_binary_operator.cc
JamesLee-Jones: Done.

PR #154, opened 2022-10-03
Rework unary operator insertion

### LINKED ISSUE #150: Why isn't unary minus inserted as part of MutationReplaceExpr?
It looks like we do not transform `e` into `-(e)`. Is that deliberate, or an omission?

@JamesLee-Jones do you recall if we discussed this?

### LINKED ISSUE #152: Missing constant-related optimisation for ~ operator
It looks like we apply `~` to the constants 0 and -1. But in the signed integer these end up being subsumed by replacing with -1 and replacing with 0.

afd:
Adds support for inserting '-'.

Avoids inserting '~' for certain constant scenarios.

Avoids inserting '!' when the argument is any integer/boolean constant.

Fixes #150.
Fixes #152.


afd:
@JamesLee-Jones for info.

PR #179, opened 2023-11-01
Implement code coverage transformation

JamesLee-Jones:
Implement a non-semantics preserving version of the code coverage instrumentation to add extra code coverage points if a mutation differs from the actual result.

### DIFF
--- a/src/dredd/src/main.cc
+++ b/src/dredd/src/main.cc
@@ -56,6 +56,13 @@ static llvm::cl::opt<bool> only_track_mutant_coverage(
                    "an input, rather than actually applying any mutants."),
     llvm::cl::cat(mutate_category));
 // NOLINTNEXTLINE
+static llvm::cl::opt<bool> semantics_preserving_coverage_instrumentation(
+        "semantics-preserving-coverage-instrumentation",
+        llvm::cl::desc("Apply a semantics preserving transformation to the source code "
+                       "which adds extra code coverage points."),
+        llvm::cl::cat(mutate_category)
+        );
+// NOLINTNEXTLINE
 static llvm::cl::opt<bool> dump_asts(
     "dump-asts",
     llvm::cl::desc("Dump each AST that is processed; useful for debugging"),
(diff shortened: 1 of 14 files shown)


JamesLee-Jones:
This is still very much a work in progress and will require large amounts of refactoring to remove code duplication. This is just a first experiment in using code coverage to guide greybox fuzzing.

JamesLee-Jones:
While trying to modify Woff2, the changes made by this update to Dredd cause compile errors:
```
SRC/src/font.cc:641:140: error: incompatible operand types ('int' and 'woff2::Font::Table *')
  if (!__dredd_enabled_mutation(9)) { return __dredd_replace_expr_bool([&]() -> bool { return static_cast<bool>(it == tables.end()); }, 0) ? __dredd_replace_expr_int(0, 3) : &it->second; }
```
```
SRC/src/woff2_enc.cc:964:91: error: comparison between pointer and integer ('const Font::Table *' and 'long')
      if (!__dredd_enabled_mutation(1475)) { if (__dredd_replace_expr_bool(table_to_store == __dredd_replace_expr_long(NULL, 1466), 1472)) table_to_store = &original; }
```
```
SRC/src/woff2_dec.cc:2191:20: error: cannot initialize a variable of type 'const woff2::Table *' with an rvalue of type 'long'
      const Table* glyf_table = __dredd_replace_expr_long(NULL, 10411);
```
Replacing `NULL` seems to be the primary culprit, but not the only one.

JamesLee-Jones:
When fuzzing `woff`, the current implementation leads to some crashes coming from the changes introduced by Dredd instead of the bug present in `woff`. This means the current implementation must not be entirely semantics preserving.

An example of a replacement that causes a crash is:
```
static unsigned int __dredd_replace_expr_unsigned_int_lvalue(unsigned int& arg, int local_mutation_id) {
  unsigned int actual_result = arg; // The crash is here
  return actual_result;
}
```

JamesLee-Jones:
https://github.com/JamesLee-Jones/dredd-fuzzing can be used for testing if the changes are semantics preserving

JamesLee-Jones:
This will fail the CI due to the cognitive complexity of the safe math checks. This could probably be solved by moving them to their own macros in the source file but isn't worth it for now.

Issue #190, opened 2023-11-17
Mutating NULL with optimisations off


JamesLee-Jones:
When optimisations are disabled, Dredd will mutate expressions such as:
```
const Table* glyf_table = NULL;
```
to become
```
const Table* glyf_table = __dredd_replace_expr_long(NULL, 10411);
```
which results in an error:
```
error: cannot initialize a variable of type 'const woff2::Table *' with an rvalue of type 'long'
```

This issue doesn't arise when optimizations are enabled, which appears to be due to the following code in `mutate_visitor.cc`:
```
    // If an expression is the direct child of a cast expression, do not mutate
    // it unless the cast is an l-value to r-value cast. In an l-value to
    // r-value cast it is worth mutating the expression before and after casting
    // since different mutations will arise. In a cast that does not change l/r-
    // value status, it is highly likely that mutations on the inner cast will
    // have the same effect as on the outer cast. The possible differences
    // arising due to a change of type are unlikely to be all that interesting,
    // and r-value to r-value implicit casts are very common, e.g. occurring
    // whenever a signed literal, such as `1`, is used in an unsigned context.
    for (const auto& parent :
         compiler_instance_->getASTContext().getParents<clang::Expr>(*expr)) {
      const auto* cast_parent = parent.get<clang::CastExpr>();
      if (cast_parent != nullptr &&
          cast_parent->isLValue() == expr->isLValue()) {
        return;
      }
    }
```
We should avoid mutating `NULL` expressions.

JamesLee-Jones:
Do you think it would be better to make this standard instead of an optimisation, or just avoid mutating NULL? @afd

JamesLee-Jones:
The following check seems to resolve the issue:
```
  if (expr->isNullPointerConstant(compiler_instance_->getASTContext(), clang::Expr::NullPointerConstantValueDependence()) != 0U) {
      return;
  }
```

afd:
@JamesLee-Jones could you attach a minimal but complete program that triggers this? Thanks.

JamesLee-Jones:
> @JamesLee-Jones could you attach a minimal but complete program that triggers this? Thanks.

For C:
```
#include <stdio.h>

int main() {
    int* x = NULL;
}
```
For C++:
```
include <cstdio>

int main() {
    int* x = NULL;
}
```

afd:
Thanks - is the include and NULL necessary? What happens if you replace NULL with 0? Or with whatever the definition of NULL is in the header?

PR #191, opened 2023-11-17
Avoid mutating NULL

JamesLee-Jones:
Avoid mutating NULL when optimizations are off as this can lead to type errors.

Fixes: #190

### DIFF
--- a/src/libdredd/src/mutate_visitor.cc
+++ b/src/libdredd/src/mutate_visitor.cc
@@ -373,16 +373,19 @@ void MutateVisitor::HandleExpr(clang::Expr* expr) {
     return;
   }
 
-  for (const auto& parent :
-       compiler_instance_->getASTContext().getParents<clang::Expr>(*expr)) {
-    const auto* cast_parent = parent.get<clang::CastExpr>();
-    if (cast_parent != nullptr &&
-        expr->isNullPointerConstant(
-            compiler_instance_->getASTContext(),
-            clang::Expr::NullPointerConstantValueDependence()) != 0U) {
-      return;
-    }
-  }
+  //  clang::Expr::NullPointerConstantKind const kind =
+  //  expr->isNullPointerConstant(
+  //      compiler_instance_->getASTContext(),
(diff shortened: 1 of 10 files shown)


JamesLee-Jones:
As discussed with @afd, a solution is to wrap the Dredd mutation in a static cast. Mutating the expressions (and NULL) is fine, we just need to make sure the return type is correct.

JamesLee-Jones:
Upon further inspection, it seems that all of the cases where this happens, the value that NULL is being compared with or assigned to is a pointer, so in this case it may not be safe to apply mutations when it is assignment and may not be useful if it is comparison.

src/libdredd/src/mutate_visitor.cc
```diff
@@ -391,6 +391,18 @@ void MutateVisitor::HandleExpr(clang::Expr* expr) {
     return;
   }
 
+  clang::Expr::NullPointerConstantKind const kind = expr->isNullPointerConstant(
```
afd, line 394: Can you add a comment for this block of code to explain what it is trying to avoid being mutated?

test/single_file/assign_null.c
```diff
@@ -0,0 +1,3 @@
+int main() {
+    int* x = 0;
```
afd, line 2: Looks like in C it is acceptable to replace a null pointer constant with the result of a function call. So we *could* make the new behaviour conditional on C++ and have a test to ensure that for C the mutation still occurs. I'm in two minds about this. On balance though I think let's keep the stricter checking that you have implemented (i.e. no action needed).

afd (changes requested):
Looks good - some small changes requested.

test/single_file/assign_null.c
JamesLee-Jones: Sounds good, I'll add a note in the comment for the changed code explaining that it would be safe to enable for C as a reminder and in case we decide to change it in the future.

afd (approved):
Looks good! Please rebase and regenerate single-file tests, then this is good to land.

PR #195, opened 2023-11-21
Make --mutation-info-file optional

JamesLee-Jones:
Make `--mutation-info-file` parameter optional as writing this file can cause issues if the mutation tree is too deep.


JamesLee-Jones:
Encapsulating whether we want to build the mutation tree in protobufs seemed like the best choice. We can't readily use `optional` as `mutation_info` isn't initialised when it is passed to `NewMutateFrontendActionFactory`.

afd (changes requested):
I didn't do a proper review yet, but at a glance I don't like the idea of adding a `use` field to the top-level protobuf object. Presumably the idea is that if this field is false / 0, the protobuf is not in use.

I think it would be better to pass an optional<MutationInfo> around, such that if this optional is empty there is no protobuf object at all, and if it's not empty there is a complete protobuf object, so that no change to the protobuf description is required.

Would that be possible or is there a good reason why the `use` approach is needed?

afd:
@JamesLee-Jones as part of this PR, can you update the project README so that the instructions no longer use this option (and also remove the comments that explain that the option is mandatory but doesn't matter for purposes of the README examples)?

afd:
@JamesLee-Jones in my latest push I updated the README so that it does use the generated .json file in the math library example, so it's only for the simple pi example that we'd want to remove all trace of it.

afd:
I now have bandwidth to look at this again if you want to rebase it and revise it if needed.

JamesLee-Jones:
> I now have bandwidth to look at this again if you want to rebase it and revise it if needed.

I'll revise it today along with the rest of the reviewed PRs.

README.md
```diff
@@ -72,11 +72,7 @@ We first show how to apply Dredd to a simple stand-alone program. We will then s
 
 ```
 # This will modify pi.cc in-place.
-# The --mutation-info-file argument is used to specify a JSON file to
-# which Dredd will output machine-readable information about the
-# mutations it applied. For the purposes of this example, this file
-# can be ignored.
-${DREDD_EXECUTABLE} examples/simple/pi.cc --mutation-info-file temp.json
+${DREDD_EXECUTABLE} examples/simple/pi.cc
```
afd, line 75: Two things:

- Below there is an `rm temp.json` that should go.

- Further down still, in the part about mutating code that uses a CMake project, the `--mutation-info-file` argument is used. Can you migrate this (currently removed) comment explaining the purpose of this argument to lower down, so that the argument is still explained when it is used?

src/libdredd/src/mutate_ast_consumer.cc
```diff
@@ -358,27 +367,33 @@ std::string MutateAstConsumer::GetDreddPreludeC(int initial_mutation_id) const {
...
+std::optional<protobufs::MutationTreeNode> MutateAstConsumer::ApplyMutations(
     const MutationTreeNode& mutation_tree_node, int initial_mutation_id,
     clang::ASTContext& context,
-    std::unordered_set<std::string>& dredd_declarations) {
+    std::unordered_set<std::string>& dredd_declarations, bool build_tree) {
   assert(!(mutation_tree_node.IsEmpty() &&
            mutation_tree_node.GetChildren().size() == 1) &&
          "The mutation tree should already be compressed.");
   protobufs::MutationTreeNode result;
+
```
afd, line 378: Looks like whitespace added by accident. (Fine to keep it if added deliberately.)

afd (changes requested):
Looking good - small changes requested.

README.md
```diff
@@ -141,6 +136,10 @@ ${DREDD_EXECUTABLE} -p build/compile_commands.json math/src/*.cc --mutation-info
 The `-p` option allows the compilation database generated by CMake above to be passed to Dredd.
 This is so that Dredd knows the correct compiler options to use when processing each source file.
 
+The --mutation-info-file argument is used to specify a JSON file to which Dredd will output
```
afd, line 139: Rephrase to "The optional --mutation-info-file argument". Also, please put `--mutation-info-file` in backticks (like I have done in this comment).

README.md
```diff
@@ -141,6 +136,10 @@ ${DREDD_EXECUTABLE} -p build/compile_commands.json math/src/*.cc --mutation-info
 The `-p` option allows the compilation database generated by CMake above to be passed to Dredd.
 This is so that Dredd knows the correct compiler options to use when processing each source file.
 
+The --mutation-info-file argument is used to specify a JSON file to which Dredd will output 
+machine-readable information about the mutations it applied. This argument is optional and for the 
+purposes of this example, `mutant-info.json` can be ignored.
```
afd, line 141: Actually it is relevant lower down - see "To learn about the mutants that Dredd has made available ...".

I suggest replacing "This argument is ... can be ignored." with "We explain below how the `mutant-info.json` file created via the `--mutation-info-file` argument can be used to query the mutants that Dredd has introduced."

afd (approved):
Thanks! A few final wording things - can you attend to them and then merge? I don't need to re-review.

afd:
@JamesLee-Jones This needs a rebase, but otherwise should be good to go if you're able to attend to the minor edits.

Issue #196, opened 2023-11-21
Conflicting types for __dredd_replace_binary_operator_LAnd_long_long_long_rhs


JamesLee-Jones:
When we have a `long long` one one side of a logical and a `long` on the other side, then the opposite later (`long` on the left, `long long` on the right), we end up with conflicting types for `__dredd_replace_binary_operator_LAnd_long_long_long_rhs` and `__dredd_replace_binary_operator_LAnd_long_long_long_lhs`.

This can be reproduced by applying Dredd to the following without optimizations and compiling:
```
int main() {
    5LL && 4L;
    4L && 5LL;
}
```
The above example isn't valid in C++.

afd:
Wow!! 🤣

afd:
I suppose we could deal with this by inserting something in the name to indicate going from one argument to the next. Or alternatively by prefixing the LHS type on the name with 1_ and the RHS type with 2_

JamesLee-Jones:
> I suppose we could deal with this by inserting something in the name to indicate going from one argument to the next. Or alternatively by prefixing the LHS type on the name with 1_ and the RHS type with 2_

That would work, we could go with arg1_..._arg2_... for clarity.

afd:
Sounds good. Are you happy to do a PR for this, adding a test exposing the issue?

PR #197, opened 2023-11-21
Add argument specifier to function name

JamesLee-Jones:
Add an argument specifier before the types in the function name. This addressed the ambiguous case where the two arguments to a binary mutation are `long long` and `long`.

Fixes: #196.


afd (approved):
LGTM.

Issue #180, opened 2023-11-01
Install cmake-format in dev_shell.sh.template


JamesLee-Jones:
It appears that `cmake-format` is not installed by `dev_shell.sh.template` which causes the `scripts/check_format.sh` script to fail.

afd:
The trouble is that cmake-format is a Python dependency, and at the moment the devshell template does't install any Python dependencies. Python can be rather fiddly - installing things with pip can work differently depending on what else you have installed, e.g. a Python virtual environment or whatever. So perhaps it's best if we address this by:

- documenting in the developer instructions that a prerequisite is having Python available and the cmake-format package of Python installed

- having the `check_format.sh` script start by issuing something like `cmake_format --version` and checking that this returns a non-zero error code. If it returns a non-zero error code a warning could be emitted that cmake format is not installed and then either check_format could error out, or instead formatting of CMake files could be skipped. (I'd be OK with the latter on the basis that this will be checked as part of CI, and it's not that common to modify CMake files.)

JamesLee-Jones:
I prefer the idea of a non-zero exit code, just so we cover the unlikely case that `cmake_format` fails to install on CI in future and we miss it.

afd:
Good point.

You happy to take this one?

JamesLee-Jones:
> Good point.
> 
> You happy to take this one?

Yeah! There is a pull request up now.

PR #202, opened 2023-11-22
Error if check-format isn't installed

JamesLee-Jones:
Check if the `check-format` python package is installed at the start of `./check_format` and return a non-zero error code. Update the documentation acordingly

Fixes: #180.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,13 @@ This project uses the [Google C++ style guide](https://google.github.io/stylegui
 
 The following guide assumes that you are using Linux for development.
 
+### Prerequisites
+
+- Python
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -243,6 +243,13 @@ This project uses the [Google C++ style guide](https://google.github.io/stylegui
 
 The following guide assumes that you are using Linux for development.
 
+### Prerequisites
+
+- Python
+- cmake-format python package (used for `./check_format`)
```
afd, line 249: I would say "(used by `scripts/check_format`)"

afd (approved):
One comment - but otherwise looks good.

PR #205, opened 2023-11-22
Refactor mutation_replace_binary_operator to use macros

JamesLee-Jones:


JamesLee-Jones:
@afd There are two approaches I could take with this PR. I could either replace the mutation of each operator with their own macro, i.e 
```
#define REPLACE_XorAssign(mutation_id_offset) if (__dredd_enabled_mutation(local_mutation_id + mutation_id_offset)) return arg1^=arg2
#define REPLACE_SubAssign(mutation_id_offset) if (__dredd_enabled_mutation(local_mutation_id + mutation_id_offset)) return arg1-=arg2
#define REPLACE_Sub(mutation_id_offset) if (__dredd_enabled_mutation(local_mutation_id + mutation_id_offset)) return arg1-arg2
#define REPLACE_ShrAssign(mutation_id_offset) if (__dredd_enabled_mutation(local_mutation_id + mutation_id_offset)) return arg1>>=arg2
#define REPLACE_ShlAssign(mutation_id_offset) if (__dredd_enabled_mutation(local_mutation_id + mutation_id_offset)) return arg1<<=arg2
```
or we instead extract the template and have one macro `MUTATE(mutation_id_offset, arg)  if (__dredd_enabled_mutation(local_mutation_id + mutation_id_offset)) return arg`.

The former approach would allow different mutation templates to be applied to different operators if desired, but the later would simplify the output code substantially. What is your opinion?

JamesLee-Jones:
The first approach is implemented at the moment, but could be changed quite easily.

afd:
I'm not sure. I think the approach with more macros, allowing more flexibility per macro, is probably better. I don't anticipate the current approach to mutation with Dredd changing much, so I guess we need to think through what your use case for coverage-guided fuzzing will need. For your use case I can see us needing quite a bit of flexibility, as each operator will need different UB checking. So it's likely the first approach you mention is better.

Note: this macros approach is just one thing we should consider. It could be that we go some way down that road and decide it's not what we want after all, but it's hard to know until we try, so thanks for initially prototyping this.

Issue #219, opened 2024-04-11
Dredd includes system headers before headers that must be before system headers


JamesLee-Jones:
This can be demonstrated by creating test.h:
```
#if defined(__GLIBC__) && !defined(__FreeBSD_kernel__) &&  defined(__CONFIG_H__)
#  error test.h must be #included before system headers
#endif
```
and test.c:
```
#include "test.h"

int main() {
  int x = 1 + 2;
}
```

This compiles with `clang test.c`, but after running `dredd test.c` the modified program no longer compiles and gives the error
```
./test.h:2:4: error: test.h must be #included before system headers
```
In more complicated projects, such as binutils, this can cause Dredd to fail to instrument any of the source files.

Issue #215, opened 2024-01-25
DREDD injects header before Feature Test Macro


JonathanFoo0523:
Minimal Reproducible Example (Produced with CReduce):
```
#define _GNU_SOURCE
#include <sys/mman.h>
void main() {
  size_t a = 4096;
  mremap;
}
```

## Environment
- OS: Ubuntu 20.04.6 LTS
- Compiler: gcc (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0
- C library: (Ubuntu GLIBC 2.31-0ubuntu9.14) 2.31

JonathanFoo0523:
Information about feature_test_macro and list of known feature_test_macro can be found 
[here](https://man7.org/linux/man-pages/man7/feature_test_macros.7.html)

PR #220, opened 2024-04-11
Insert Dredd declartions before first function

JamesLee-Jones:
Inserting Dredd's declarations at the beginning of the file can cause
issues with header files that have to be before any system headers. So
instead, insert Dredd's declarations at the top level before the first
mutation that we find.

Fixes: #219, #215

### DIFF
--- a/src/libdredd/include_private/include/libdredd/mutate_visitor.h
+++ b/src/libdredd/include_private/include/libdredd/mutate_visitor.h
@@ -179,9 +179,12 @@ class MutateVisitor : public clang::RecursiveASTVisitor<MutateVisitor> {
   // this special case, so that it can be ignored.
   bool IsConversionOfEnumToConstructor(const clang::Expr& expr) const;
 
-  // It is safe to Dredd's prelude before the first function we encounter in a
-  // file as Dredd only makes source code modifications inside functions.
+  // It is safe to put Dredd's prelude before the first function we encounter in
+  // a file as Dredd only makes source code modifications inside functions.
   void UpdateStartLocationOfFirstFunctionInSourceFile();
+
+  // TODO(JLJ): Add comment
+  void AddMutation(std::unique_ptr<Mutation> mutation);
 
   const clang::CompilerInstance* compiler_instance_;
   bool optimise_mutations_;
--- a/src/libdredd/include_private/include/libdredd/mutate_visitor.h
+++ b/src/libdredd/include_private/include/libdredd/mutate_visitor.h
@@ -179,6 +179,8 @@ class MutateVisitor : public clang::RecursiveASTVisitor<MutateVisitor> {
   // this special case, so that it can be ignored.
   bool IsConversionOfEnumToConstructor(const clang::Expr& expr) const;
 
+  // It is safe to Dredd's prelude before the first function we encounter in a
+  // file as Dredd only makes source code modifications inside functions.
   void UpdateStartLocationOfFirstFunctionInSourceFile();
 
   const clang::CompilerInstance* compiler_instance_;
--- a/src/libdredd/include_private/include/libdredd/mutate_visitor.h
+++ b/src/libdredd/include_private/include/libdredd/mutate_visitor.h
@@ -108,6 +108,11 @@ class MutateVisitor : public clang::RecursiveASTVisitor<MutateVisitor> {
     return constant_sized_arrays_to_rewrite_;
   }
 
+  [[nodiscard]] clang::SourceLocation
+  GetStartLocationOfFirstFunctionInSourceFile() const {
+    return start_location_of_first_function_in_source_file_;
+  }
+
  private:
   // Helper class that uses the RAII pattern to support pushing a new mutation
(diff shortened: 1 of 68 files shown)


JamesLee-Jones:
Although this is unlikely to ever be unreasonably expensive, there is probably an optimizations that could be made. If the Clang AST is always traversed top down, then we just need to find the place that Dredd is going to modify and insert the prelude before this at the global scope. This means once `start_location_of_first_function_in_source_file_` is valid, we could skip the rest of the function as this will always have the earliest starting source location.

src/libdredd/src/mutate_visitor.cc
```diff
@@ -63,6 +64,25 @@ bool MutateVisitor::IsTypeSupported(const clang::QualType qual_type) {
...
+      auto source_range_in_main_file = GetSourceRangeInMainFile(
+          compiler_instance_->getPreprocessor(), *enclosing_decls_[0]);
+      if (start_location_of_first_function_in_source_file_.isInvalid() ||
+          comparator(source_range_in_main_file.getBegin(),
+                     start_location_of_first_function_in_source_file_)) {
+        start_location_of_first_function_in_source_file_ =
+            source_range_in_main_file.getBegin();
+      }
+    }
+  }
```
JamesLee-Jones, line 83: Use `IsInFunction()` rather than repeating.

afd: @JamesLee-Jones I didn't fully understand this comment and am not sure whether this is a remaining TODO, or something to be addressed in a follow up PR. If there is a way to simplify this then please go ahead.

afd:
@JamesLee-Jones Can you rebase this one please? From what I recall it should be good to go, except there was one simplification you were looking at making (see open discussion).

src/libdredd/include_private/include/libdredd/mutate_visitor.h
```diff
@@ -167,9 +172,17 @@ class MutateVisitor : public clang::RecursiveASTVisitor<MutateVisitor> {
   // this special case, so that it can be ignored.
   bool IsConversionOfEnumToConstructor(const clang::Expr& expr) const;
 
+  // It is safe to Dredd's prelude before the first function we encounter in a
```
afd, line 175: Missing word after "to".

src/libdredd/src/mutate_visitor.cc
```diff
@@ -508,6 +524,7 @@ bool MutateVisitor::VisitExpr(clang::Expr* expr) {
     return true;
   }
 
+  UpdateStartLocationOfFirstFunctionInSourceFile();
```
afd, line 527: I think it would be cleaner to call this whenever a mutation is added.

I propose:

- introduce a private helper function, `AddMutation`
- have the body of `AddMutation` do the `mutation_tree_path_.back()->AddMutation(...)` call that is currently made a few times in `mutate_visitor.cc`, and replace those calls with a call to the new `AddMutation` method
- Have the new `AddMutation` method call `UpdateStartLocationOfFirstFunctionInSourceFile` as the first thing that it does.

I think this would be good because if in the future we add a new kind of mutation to Dredd, we won't need to remember to add calls to `UpdateStartLocation...`.

src/libdredd/src/mutate_visitor.cc
```diff
@@ -571,6 +588,7 @@ bool MutateVisitor::TraverseCompoundStmt(clang::CompoundStmt* compound_stmt) {
     assert(!enclosing_decls_.empty() &&
            "Statements can only be removed if they are nested in some "
            "declaration.");
+    UpdateStartLocationOfFirstFunctionInSourceFile();
```
afd, line 591: See previous comment.

afd (changes requested):
One typo, and a suggestion for a small refactoring that I think will be worthwhile.

afd:
@JamesLee-Jones This will need a rebase. Does my suggested refactoring make sense and are you good to do it?

JamesLee-Jones:
> @JamesLee-Jones This will need a rebase. Does my suggested refactoring make sense and are you good to do it?

Your suggestion makes sense, I'll refactor it today!

src/libdredd/include_private/include/libdredd/mutate_visitor.h
```diff
@@ -174,9 +180,20 @@ class MutateVisitor : public clang::RecursiveASTVisitor<MutateVisitor> {
   // this special case, so that it can be ignored.
   bool IsConversionOfEnumToConstructor(const clang::Expr& expr) const;
 
+  // It is safe to put Dredd's prelude before the first function we encounter in
+  // a file as Dredd only makes source code modifications inside functions.
+  void UpdateStartLocationOfFirstFunctionInSourceFile();
+
+  // TODO(JLJ): Add comment
```
afd, line 187: How about:

> Adds details of a mutation that can be applied, and performs associated bookkeeping.

afd (approved):
Thanks! Good to go after making some very final changes.

Issue #222, opened 2024-05-21
Incorrect validation for binary arithmetic redundant condition


khemichew:
**Problem**
Dredd incorrectly checks the right operand twice instead of both the left and right operand when omitting redundant mutant generation:

https://github.com/mc-imperial/dredd/blob/11dac1f376eb960cf6dda4f4b04cd00bf69f3ebe/src/libdredd/src/mutation_replace_binary_operator.cc#L1058-L1065

**Fix**
The first half of the validation condition should be:

```cpp
(MutationReplaceExpr::ExprIsEquivalentToInt(*binary_operator_->getLHS(), 
                                                 0, ast_context) || 
      MutationReplaceExpr::ExprIsEquivalentToFloat(*binary_operator_->getLHS(), 
                                                   0.0, ast_context))
```

Issue #225, opened 2024-05-22
Incorrect transformation of braced initialization


JamesLee-Jones:
The expression:
``` char test {24}; ```
is incorrectly translated to:
```
char test __dredd_replace_expr_char_constant(__dredd_replace_expr_char_constant({__dredd_replace_expr_char_constant(24, 0)}, 5), 10);
```

The literal in the braces should continue being mutated, but we should avoid mutating expressions if they are being used to initialise a variable.

PR #226, opened 2024-05-22
Fix braced initalization and binary zero optimizations

JamesLee-Jones:
Previously, when mutating expressions such as `char test {24}`, the braced initialization expression `{24}` would also get mutated (as noted in #225). This lead to compilation errors. This is resolved by avoiding the mutation of `InitListExpr`.

Also, the optimization checks for the case when both arguments to a binary expression were zero was previously wrong (as noted in #222). This is resolved.

Fixes: #222, #225.


JamesLee-Jones:
This appears to only be failing because of the CI as noted in #227.

afd (approved):

PR #231, opened 2024-05-24
Attempt to fix issue 230

JonathanFoo0523:
The `GetRegularDreddPreludeCpp` and `GetRegularDreddPreludeC` functions have been modified with the following changes:

(a) The return type is now `bool` instead of `int`. If the return expression is of type `uint64_t`, returning an `int` could truncate the value. By returning a `bool`, the result will be `0` if the value compares equal to `0`; otherwise, the result will be `1`.

(b) Any `int` literal involved in a bit shift operation is explicitly cast to `uint64_t`.

Additionally, we have added a new directory at `test/compile_execute_files`, which contains the C/C++ program, expected std output, and runner script for simple regression tests.


JamesLee-Jones:
Provided @afd is happy with it, I can keep running the CI but it's useful to run `scripts/check_all.sh` as well as this will give you a good idea of whether it will pass.

afd (changes requested):
Thanks for this! Can you please:

- Put your changes on a branch under `JonoathanFoo0523` - this will allow me to check out that branch to review your changes locally.

- Adapt the title of the PR to indicate what it's about. It is good practice to mention the issue in the PR, but you usually do that in the PR description, e.g. via "Fixes #231." In the title of the PR it's good to use words that relate to the content of the contribution, so that when developers look through a list of PRs they can tell roughly what they are about.

- Also, since you're going to put this on a branch, etc., please rebase your changes over the latest changes to main, and re-generate the single file test cases (as other commits have added a few more in the meantime).

src/libdredd/src/mutate_ast_consumer.cc
```diff
@@ -284,7 +285,7 @@ std::string MutateAstConsumer::GetRegularDreddPreludeC(
   result << "#endif\n";
   result << "\n";
   result << "static thread_local int __dredd_some_mutation_enabled = 1;\n";
-  result << "static int __dredd_enabled_mutation(int local_mutation_id) {\n";
+  result << "static bool __dredd_enabled_mutation(int local_mutation_id) {\n";
```
afd, line 288: I'm fine with using `bool` here as it is more readable, but because C does not have a `bool` type, I think this will end up being the same as `int`. Therefore this change might not have any impact. Again - I am fine with the change, but wanted to bring this up in case your reasoning was that it would have impact.

test/compile_execute_file/bit_overflow.c.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
afd, line 1: It is great that you have these scripts. Now what we need is a way to include them in the CI.

Please can you have a look at:

- `scripts/check_single_file_tests.sh`
- `check_all.sh`

I wonder if you might make something called:

- `scripts/check_compile_execute_file.sh`

This should run all of the `.sh` scripts of the form you have added (there are only two now, but we may add more later).

Then update the `check_all.sh` script so that it runs the new `check_compile_execute_file.sh` script.

That should ensure that this all runs in the CI.

afd (changes requested):
A couple of TODOs for @JonathanFoo0523

test/compile_execute_file/bit_overflow.c.sh
JamesLee-Jones: It's probably worth compiling the examples with expected output with `-fsanitize=undefined` as I believe this would have also caught this error and if no mutants are enabled, Dredd shouldn't introduce any undefined behavior.

JonathanFoo0523:
> Thanks for this! Can you please:
> 
> * Put your changes on a branch under `JonoathanFoo0523` - this will allow me to check out that branch to review your changes locally.
> * Adapt the title of the PR to indicate what it's about. It is good practice to mention the issue in the PR, but you usually do that in the PR description, e.g. via "Fixes [Attempt to fix issue 230 #231](https://github.com/mc-imperial/dredd/pull/231)." In the title of the PR it's good to use words that relate to the content of the contribution, so that when developers look through a list of PRs they can tell roughly what they are about.
> * Also, since you're going to put this on a branch, etc., please rebase your changes over the latest changes to main, and re-generate the single file test cases (as other commits have added a few more in the meantime).

I don't think GitHub allows me to change the source branch (from JonathanFoo0523
) of a PR to different branch (e.g., JonathanFoo0523
). If necessary, I can close this PR and create a new PR with different source branch.

test/compile_execute_file/bit_overflow.c.sh
afd: @JamesLee-Jones Good idea.

afd:
@JonathanFoo0523 Yes, please open a new PR with this on a branch.

test/compile_execute_file/bit_overflow.c.sh
JonathanFoo0523: @afd I saw you are implementing some kind of CI test at [PR 255](https://github.com/mc-imperial/dredd/pull/255). Does it overlap with `check_compile_execute_file.sh` mentioned here?

afd: Yes, I think it probably will. I should get #255 landed soon. Once it's done, perhaps you can rework your PR to use the same infrastructure that I am bringing in there.

afd:
@JonathanFoo0523 When you do your new PR for this, please can you see if you can set up the new test case as an "execute" test, using the infrastructure I have added via #255 ? Please see the README under `tests/execute`, as well as the `tests/execute/add` example - hopefully this can be made to fit the bill for what you are doing.

JonathanFoo0523:
@afd Sure! I am working on it.

Issue #233, opened 2024-05-25
Dredd mutates constexpr


JamesLee-Jones:
Dredd currently mutates `constexpr` expressions. Since these should be possible to evaluate at compile time, this is not valid unless the Dredd mutation functions are made `constexpr` accordingly or we avoid mutating them.

The following example demonstrates this error.
```
#include <iostream>

constexpr int Max(int a, int b) { return a > b ? a : b }

int main() {
  std::cout << Max(5,4) << std::endl;
}
```
This gives the following error upon compilation:
```
test.cc:111:15: error: constexpr function never produces a constant expression [-Winvalid-constexpr]
constexpr int Max(int a, int b) { if (!__dredd_enabled_mutation(27)) { return __dredd_replace_expr_int(__dredd_replace_expr_int_lvalue(__dredd_replace_expr_bool_omit_true(__dredd_replace_binary_operator_GT_arg1_int_arg2_int(__dredd_replace_expr_int(__dredd_replace_expr_int_lvalue(a, 0), 2) , __dredd_replace_expr_int(__dredd_replace_expr_int_lvalue(b, 8), 10), 16), 18) ? a : b, 19), 21); } }
              ^
test.cc:111:40: note: non-constexpr function '__dredd_enabled_mutation' cannot be used in a constant expression
constexpr int Max(int a, int b) { if (!__dredd_enabled_mutation(27)) { return __dredd_replace_expr_int(__dredd_replace_expr_int_lvalue(__dredd_replace_expr_bool_omit_true(__dredd_replace_binary_operator_GT_arg1_int_arg2_int(__dredd_replace_expr_int(__dredd_replace_expr_int_lvalue(a, 0), 2) , __dredd_replace_expr_int(__dredd_replace_expr_int_lvalue(b, 8), 10), 16), 18) ? a : b, 19), 21); } }
                                       ^
test.cc:16:13: note: declared here
static bool __dredd_enabled_mutation(int local_mutation_id) {
            ^
1 error generated.

```

JamesLee-Jones:
Since we can't tell at compile time whether a mutation is enabled, I suppose we will have to avoid mutating these expressions. The approach to this may be different in #179.

PR #234, opened 2024-05-25
Avoid mutation constexpr specified functions

JamesLee-Jones:
We cannot mutate functions that are specified as constexpr as Dredd's mutations occur dynamically and constexpr functions require compile time evaluation.

Fixes: #233.

### DIFF
--- a/src/libdredd/src/mutate_visitor.cc
+++ b/src/libdredd/src/mutate_visitor.cc
@@ -117,6 +117,13 @@ bool MutateVisitor::TraverseDecl(clang::Decl* decl) {
     // associated with static assertions anyway.
     return true;
   }
+  if (const auto* function_decl = llvm::dyn_cast<clang::FunctionDecl>(decl)) {
+    if (function_decl->isConstexpr()) {
+      // Because Dredd's mutations occur dynamically, they cannot be applied to
+      // C++ constexpr functions, which require compile-time evaluation.
+      return true;
+    }
+  }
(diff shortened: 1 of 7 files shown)


afd (commented):
This is interesting! I am pretty sure your change is good, but I'm curious: if we had this code:

```
constexpr int Max(int a, int b) { return a > b ? a : b; }

int foo() {
  return Max(5,4);
}
```

then would we expect the call to "Max(5,4)" to get mutated, because although it is a constexpr, it is not being returned in a constexpr context?

Would you be able to have a look at that, and if the call to Max does get mutated add another single file test case to demonstrate it?

JamesLee-Jones:
> then would we expect the call to "Max(5,4)" to get mutated, because although it is a constexpr, it is not being returned in a constexpr context?
> 
> Would you be able to have a look at that, and if the call to Max does get mutated add another single file test case to demonstrate it?

In this case, the call to Max does get mutated as we would expect, producing:
```
# Dredd prelude

constexpr int Max(int a, int b) { return a > b ? a : b; }

int foo() {
  if (!__dredd_enabled_mutation(15)) { return __dredd_replace_expr_int_constant([&]() -> int { return Max(__dredd_replace_expr_int_constant(5, 0),__dredd_replace_expr_int_constant(4, 5)); }, 10); }
}
```

So I'll add another test case.

afd (approved):
Thanks for adding the extra test case - this is good to go!

Issue #235, opened 2024-05-25
Dredd mutates variables that should have a constant initializer


JamesLee-Jones:
Dredd still mutates some forms of variables that should have a constant initializer. This is demonstrated when applying Dredd to the following example:
```
int main() {
  [[clang::require_constant_initialization]] static int x = __dredd_replace_expr_int_zero(0, 0);
}
```
This gives:
```
// Dredd Prelude ...
int main() {
  [[clang::require_constant_initialization]] static int x = __dredd_replace_expr_int_zero(0, 0);
}
```

Although Dredd doesn't currently support C++ 20, I suspect the same issue would arise with `constinit`.

PR #236, opened 2024-05-25
Avoid mutating variables that require const initialization

JamesLee-Jones:
We cannot mutate variable declarations that require constant initialization as Dredd's mutations occur dynamically.

Fixes: #235.

### DIFF
--- a/src/libdredd/src/mutate_visitor.cc
+++ b/src/libdredd/src/mutate_visitor.cc
@@ -123,7 +123,7 @@ bool MutateVisitor::TraverseDecl(clang::Decl* decl) {
     }
   }
   if (const auto* var_decl = llvm::dyn_cast<clang::VarDecl>(decl)) {
-    if (var_decl->isConstexpr() || var_decl->hasConstantInitialization()) {
+    if (var_decl->isConstexpr() || var_decl->hasAttr<clang::ConstInitAttr>()) {
       // Because Dredd's mutations occur dynamically, they cannot be applied to
       // C++ constexprs or variables that require constant initialization as
       // these both require compile-time evaluation.
--- a/src/libdredd/src/mutate_visitor.cc
+++ b/src/libdredd/src/mutate_visitor.cc
@@ -123,7 +123,7 @@ bool MutateVisitor::TraverseDecl(clang::Decl* decl) {
     }
   }
   if (const auto* var_decl = llvm::dyn_cast<clang::VarDecl>(decl)) {
-    if (var_decl->isConstexpr()) {
+    if (var_decl->isConstexpr() || var_decl->hasConstantInitialization()) {
       // Because Dredd's mutations occur dynamically, they cannot be applied to
       // C++ constexprs, which require compile-time evaluation.
       return true;
--- a/src/libdredd/src/mutate_visitor.cc
+++ b/src/libdredd/src/mutate_visitor.cc
@@ -125,7 +125,8 @@ bool MutateVisitor::TraverseDecl(clang::Decl* decl) {
   if (const auto* var_decl = llvm::dyn_cast<clang::VarDecl>(decl)) {
     if (var_decl->isConstexpr() || var_decl->hasConstantInitialization()) {
       // Because Dredd's mutations occur dynamically, they cannot be applied to
-      // C++ constexprs, which require compile-time evaluation.
+      // C++ constexprs or variables that require constant initialization as
+      // these both require compile-time evaluation.
       return true;
     }
     if (!compiler_instance_->getLangOpts().CPlusPlus &&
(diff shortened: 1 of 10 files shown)


src/libdredd/src/mutate_visitor.cc
```diff
@@ -118,7 +118,7 @@ bool MutateVisitor::TraverseDecl(clang::Decl* decl) {
     return true;
   }
   if (const auto* var_decl = llvm::dyn_cast<clang::VarDecl>(decl)) {
-    if (var_decl->isConstexpr()) {
+    if (var_decl->isConstexpr() || var_decl->hasConstantInitialization()) {
       // Because Dredd's mutations occur dynamically, they cannot be applied to
       // C++ constexprs, which require compile-time evaluation.
```
afd, line 123: Can you please update the comment to reflect the additional check? I hadn't heard of this, but from [here](https://clang.llvm.org/docs/AttributeReference.html#require-constant-initialization-constinit-c-20
) it looks like `require_constant_initialization` attributes are supported by GCC and Clang, and that C++20 adds a `constinit` keyword.

Could you also add some tests that use the `constinit` keyword instead of the attribute? (Hopefully we can support C++20 tests in Dredd's single file tests.)

afd (changes requested):
This looks great - a minor comment change and (if possible) some extra tests requested.

src/libdredd/src/mutate_visitor.cc
JamesLee-Jones: Unfortunately because Dredd uses an earlier version of clang and the C++ standard, it throws the following error when trying to mutate a variable declaration using `constinit`:
```
/home/jlj/dev/dredd/temp/const_init_keyword.cc:2:3: error: unknown type name 'constinit'
  constinit static int x = 0;

```
This definitely feels like an issue to keep for future when we switch to using C++ 20.

afd:
@JamesLee-Jones This fails with:

```
2024-07-05T13:45:23.8860447Z + echo 'Found 67841 mutants when mutating the SPIR-V validator source code. Expected 67998. If Dredd changed recently, the expected value may just need to be updated, if it still looks sensible. Otherwise, there is likely a problem.'
2024-07-05T13:45:23.8861595Z + exit 1
2024-07-05T13:45:23.8862744Z Found 67841 mutants when mutating the SPIR-V validator source code. Expected 67998. If Dredd changed recently, the expected value may just need to be updated, if it still looks sensible. Otherwise, there is likely a problem.
```

It therefore looks as if your change is slightly reducing the number of mutants that are available when spirv-val is mutated. We should look into this. The approach would be:

- mutate spirv-val just like the CI does with a version of dredd before this change
- mutate a different checkout of spirv-val with dredd after this change
- diff the mutated code - this should shed light on which file contains the code that exhibits the discrepancy

After that some manual or automated test case reduction should home in on where the problem is.

JamesLee-Jones:
> @JamesLee-Jones This fails with:
> 
> ```
> 2024-07-05T13:45:23.8860447Z + echo 'Found 67841 mutants when mutating the SPIR-V validator source code. Expected 67998. If Dredd changed recently, the expected value may just need to be updated, if it still looks sensible. Otherwise, there is likely a problem.'
> 2024-07-05T13:45:23.8861595Z + exit 1
> 2024-07-05T13:45:23.8862744Z Found 67841 mutants when mutating the SPIR-V validator source code. Expected 67998. If Dredd changed recently, the expected value may just need to be updated, if it still looks sensible. Otherwise, there is likely a problem.
> ```
> 
> It therefore looks as if your change is slightly reducing the number of mutants that are available when spirv-val is mutated. We should look into this. The approach would be:
> 
>     * mutate spirv-val just like the CI does with a version of dredd before this change
> 
>     * mutate a different checkout of spirv-val with dredd after this change
> 
>     * diff the mutated code - this should shed light on which file contains the code that exhibits the discrepancy
> 
> 
> After that some manual or automated test case reduction should home in on where the problem is.

From looking at the single file test that was modified, I believe it is because the check i used catches `const` marked variables as well as those marked as requiring const initialization and thus no mutants are produced for expressions of the form `const type = value;`.

afd:
@JamesLee-Jones can you rebase this, please?

@JonathanFoo0523 after James has rebased, can you investigate why James's change is reducing the number of mutants that get created for spirv-val? (See cxx_apps.sh under `.github/workflows` for details of how spirv-val is handled.)

JamesLee-Jones:
> @JamesLee-Jones can you rebase this, please?
> 
> @JonathanFoo0523 after James has rebased, can you investigate why James's change is reducing the number of mutants that get created for spirv-val? (See cxx_apps.sh under `.github/workflows` for details of how spirv-val is handled.)

I believe [this commit](https://github.com/mc-imperial/dredd/pull/236/commits/62514743ede3c2bffaa5270248d1f97870ffb65c) already resolved the issue.

afd (approved):
LGTM!

PR #238, opened 2024-05-29
Fix incorrect mutation of comma expressions in C++

JamesLee-Jones:


src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -538,18 +538,18 @@ void MutationReplaceExpr::ReplaceExprWithFunctionCall(
...
-      }
+  if (!ast_context.getLangOpts().CPlusPlus && expr_->isLValue() &&
+      input_type.ends_with('*')) {
+    prefix.append("&(");
+    suffix = ")" + suffix;
+  }
+
+  if (const auto* binary_expr = llvm::dyn_cast<clang::BinaryOperator>(expr_)) {
+    // The comma operator requires special care in C, to avoid it appearing to
+    // provide multiple parameters for an enclosing function call.
```
afd, line 549: Can you remove the comment about "in C"?

I think that initially, because we did everything using lambdas for C++, this was not a problem for C++. But now that we try to avoid using lambdas where possible, I see that it's also a problem for C++.

src/libdredd/src/mutation_replace_expr.cc
```diff
@@ -538,18 +538,18 @@ void MutationReplaceExpr::ReplaceExprWithFunctionCall(
...
+    if (binary_expr->isCommaOp()) {
+      prefix.append("(");
+      suffix = ")" + suffix;
```
afd, line 552: I'm sure you're right with this change. It's not all that obvious to me, but I'm not "in the zone" with this at present. I see you have lots of tests, which is great.

afd (changes requested):
This looks good - a very small change to a comment requested.

src/libdredd/src/mutation_replace_expr.cc
JamesLee-Jones: I seem to remember that it's the case that something being a C LValue is not a mutually exclusive check from wrapping a comma operator with brackets. Hence the issue occurring when initializing variables using a comma operator.

afd (approved):
LGTM!

PR #255, opened 2024-07-08
Add `execute` tests

### LINKED ISSUE #254: Dredd tests that check functional correctness of mutations
Some of the optimisations that Dredd does to avoid redundant or equivalent mutants are fiddly. It would be good to have at least some exhaustive tests to check for these. What I'm thinking about is something like:

- a single-file example
- an "expectations" file, comprising a series of Python regular expressions, one per line

For each such test, the following should be done:

- Dredd is applied to the single-file example
- The example is compiled with sanitizers (ubsan and asan)
- The example is executed with no mutations enabled, and it is confirmed that its standard output matches the first regex
- The example is executed again and again with each possible mutant, and it is confirmed that:
    - For each mutant, at least one regex is matched, either by standard output or standard error (the latter so that mutants that will lead to UB can be checked) 
    - All the regexes are matched at least once, across all of the mutants

This will allow us to check, for example, that optimisations are not leading to a loss of coverage.

afd:
Adds infrustructure and one example of a new kind of `execute` test, that allows checking of the various program outputs that should be expected when all mutations for a file are considered.

Fixes #254.


test/execute/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+There is one subdirectory per test.
+
+Each subdirectory must contain:
+
+- `harness.c(c)` - a C or C++ file with a `main`, that should print a single line of output.
+
+- `tomutate.c(c)` - a C or C++ file that will be mutated. The harness should call functions of this file.
+
+- `original.txt` - A file containing the single line of output that should be printed by the harness when no mutants are enabled.
```
JonathanFoo0523, line 12: Is "single line"  required?

I noted that the the check is performed by extracting output
```
original_result.stdout.decode('utf-8')
```
in `check_one_execute_test.py`. which should works for multiple line?

JonathanFoo0523: Moreover, the expected output is extracted by
```
open(test_directory / 'original.txt', 'r').read()
```
which should read more than a line.

afd:
Hi Jonathan

You are right that at present the output checking would support multi-line output. However, the checking for mutant output is geared towards one line of output per mutant, so I think it's best if we stick with tests that produce just one line of output.

Feel to do a PR adding better checks for this in the Python file if you like - it would be better if a test that produces multiple lines of output (when no mutants are enabled) would fail.

Thanks

Ally

________________________________

PR #274, opened 2024-07-12
Trivial change.

afd:


afd:
This has confirmed that nothing has broken behind the scenes with Windows / Mac builds (#273 is failing mysteriously for those platforms).

PR #278, opened 2024-07-12
Prevent expression-under-cast-bypass in Initializer Lists for more mutations and narrowing type handling

### LINKED ISSUE #270: Mutation of expression used in initializer list results in ill-formed program
Applying Dredd to
```
#include <initializer_list>
#include <cstdint>

template <typename T> class foo {
public:
  foo(std::initializer_list<T>) {}
};

int main() { 
    foo<uint64_t>{+2}; 
}
```
change the sole statement in `main()` to 
```
if (!__dredd_enabled_mutation(5)) { foo<uint64_t>{+__dredd_replace_expr_int_constant(2, 0)}; } 
```

This causes the following compilation error:
```
non-constant-expression cannot be narrowed from type 'int' to 'unsigned long' in initializer list
```

JonathanFoo0523:
Avoid bypass of implicit inner cast mutation under `default` (optimize_mutations) mode for expressions in Initializer List. Besides producing more mutants, this change ensures that Dredd can correctly handle narrowing type expressions involving `std::initializer_list` without bypassing necessary `static_cast` operations.

Fixes #270

### DIFF
--- a/.github/workflows/cxx_apps.sh
+++ b/.github/workflows/cxx_apps.sh
@@ -157,7 +157,7 @@ pushd SPIRV-Tools
   ./build/test/val/test_val_fghijklmnop
   ./build/test/val/test_val_rstuvw
   NUM_MUTANTS=`python3 ${DREDD_ROOT}/scripts/query_mutant_info.py mutation-info.json --largest-mutant-id`
-  EXPECTED_NUM_MUTANTS=68821
+  EXPECTED_NUM_MUTANTS=68841
   if [ ${NUM_MUTANTS} -ne ${EXPECTED_NUM_MUTANTS} ]
   then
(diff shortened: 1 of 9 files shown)


JonathanFoo0523:
Check failed due to
```
Found 68841 mutants when mutating the SPIR-V validator source code. Expected 68821. If Dredd changed recently, the expected value may just need to be updated, if it still looks sensible. Otherwise, there is likely a problem.
```

afd:
Oh, looks like you already did this for SPIR-V, so you probably just need to do it for LLVM. You'll have to download the log files to see the change that's needed, as the CI output for cxx_apps is very large.

afd (approved):
LGTM - please merge once bots pass.

afd:
This is failing because the cognitive complexity of `HandleExpr` has gotten too high. This is normally where one needs to think about extracting out a helper function. Is there an obvious piece of logic related to your additions that could be extracted out?

PR #279, opened 2024-07-13
Prevent mutation on cast of passed-by-reference bitfield l-value 

### LINKED ISSUE #259: Dredd incorrectly apply mutation to lvalue inside a function that expect lvalue
Applying dredd to the following file
```
#include <tuple>
struct foo {
  int b : 2;
};

int main() {
  const foo d = foo();
  std::tie(d.b);
}
```
results in
```
#include <tuple>
struct foo {
  int b : 2;
};

void c() {
  const foo d = foo();
  if (!__dredd_enabled_mutation(2)) { std::tie(__dredd_replace_expr_int_zero(d.b, 0)); }
}
```
which, when compile, results in error related to 
```
no matching function for call to 'tie': not viable: expects an lvalue for 1st argument
```

JonathanFoo0523:
Addresses an issue where implicit casts on bitfield l-values, subsequently passed by reference, are mutated. To prevent this, we check if the underlying value is a bitfield l-value and if the expression is a child of `MaterializeTemporaryExpr`. If both conditions are met, the mutation is avoided.

Fixes #259


afd (approved):
Very nice - thanks for this!

PR #281, opened 2024-07-13
Rework protobuf representation of mutation tree

### LINKED ISSUE #192: Error writing JSON data when input file is large
When applying Dredd to a very large source file (one generated by CSmith for example), calling `MessageToJsonString` in `main.cc` results in the following json generation status:
```
INVALID_ARGUMENT:Message too deep. Max recursion depth reached for type 'dredd.protobufs.SourceLocation', field 'start'
```
This is currently hidden as the only error returned is "Error writing JSON data to {filename}".

We should make the error messages clearer for cases that are expected to error, and should either limit the input size to prevent this happening or find a way to output the json anyway.

afd:
To avoid a restriction of the protobuf library that limits the
recursion depth of a protobuf message, this change reworks the way the
mutation tree is represented in protobuf format. The tree is now
organised as a flat list of nodes, and the children of a node are
identified via indices into this list.
    
Fixes #192.


afd:
@JamesLee-Jones This reworking turned out to be reasonably straightforward, so - assuming that this solves the problems that you and @JonathanFoo0523 encountered - I think it's better to go this route rather than trying to upstream work-arounds to protobufs.

JonathanFoo0523 (commented):
LGTM

PR #284, opened 2024-07-13
Add execute test to check unsigned comparisons

afd:
This currently fails due to what must be a bug in optimisation of mutations.


afd:
Split this into #291 and #292.

PR #287, opened 2024-07-13
Rewrite size expressions of constant-sized constexpr arrays

### LINKED ISSUE #283: Dredd failed to rewrite sizes of constant-sized arrays with `constexpr` qualifier
Applying dredd to
```
int main() {
  const int a = 4;
  static constexpr char b[a]{};
}
```
produce
```
int main() {
  const int a = __dredd_replace_expr_int_constant(4, 0);
  static constexpr char b[a]{};
}
```
which caused the error
```
error: variable length array declaration cannot have 'static' storage duration
```

This bug is similar in nature to issue #263 (addressed by PR #273 )

afd:
Enhances the support for rewriting the size expressions of constant-sized arrays so that arrays with the constexpr qualifier are also handled.

Fixes #283.


JonathanFoo0523 (approved):
LGTM

JonathanFoo0523:
Merged!

PR #293, opened 2024-07-15
Rewrite array's size in `struct`, and argument of `static_assert` and `__builtin_frame_address`

### LINKED ISSUE #286: Dredd mutates `const` variable used in `static_assert`
Dredd mutate the C++ program
```
int main() {
  const int a = 1;
  static_assert(a);
}
```
to
```
int main() {
  const int a = __dredd_replace_expr_int(1, 0);
  static_assert(a);
}
```
which lead to compilation error:
```
error: static assertion expression is not an integral constant expression
```

### LINKED ISSUE #289: Dredd mutate expression in `__builtin_frame_address()`
Dredd mutates the `__builtin_frame_address()` in 
```
void foo() {
    __builtin_frame_address(0);
}
```
to
```
__builtin_frame_address(__dredd_replace_expr_unsigned_int_zero(0, 0));
```
leading to compilation error:
```
argument to '__builtin_frame_address' must be a constant integer
```

### LINKED ISSUE #290: Dredd failed to rewrite const array size in struct
Applying dredd to
```
int main() {
  const int a = 0;
  struct b {
    int bar[a];
  };
}
```
results in 
```
int main() {
  const int a = __dredd_replace_expr_int_zero(0, 0);
  struct b {
    int bar[a];
  };
}
```
This results in the following compilation error
```
error: fields must have a constant size: 'variable length array in structure' extension will never be supported
```

JonathanFoo0523:
Enhances the support for rewriting the size expressions of
constant-sized arrays in `struct` fields. Additionally, rewrites the
arguments' expressions of `static_assert()` and
`__builtin_frame_address()`, which need to be compile-time constants,
while still allowing the constants that make up the expression— which
may also be referred to elsewhere— to be freely mutated.

Fixes #286 , Fixes #289 , Fixes #290


JonathanFoo0523:
Failed on window because:
```
'__builtin_frame_address': identifier not found
```

The corresponding function for window is `_AddressOfReturnAddress` and program made use of directive to work on both system. Eg:
```
void foo() {
#if __GNUC__ || __has_builtin(__builtin_frame_address)
    return __builtin_frame_address(0);
#elif defined(_MSC_VER)
    return _AddressOfReturnAddress();
#endif
}
```

But this can't be used as test case, because different platform produce different dredd-mutated program. 

@afd Any suggestion to solve this?

afd:
Please see `build.sh`, under `.github/workflows`. You'll see there that there are exclusions that remove certain test cases that are known not to work under Windows. I suggest you make this a Linux-only test case, and add it to that exclusion list.

It would be neat to have this tested with the Windows variant too, but it's not a priority so I wouldn't worry about it.

src/libdredd/src/mutate_visitor.cc
```diff
@@ -108,14 +108,23 @@ bool MutateVisitor::TraverseDecl(clang::Decl* decl) {
...
-    return true;
+
+    // However, some C++ compiler complain if the expression under static
+    // assertion is not a compile-time constants. To allow the constant
+    // declaration that occurs inside static assertion to have their initial
+    // value mutated (because they may be used elsewhere), but to avoid
+    // compilaton errors, we record the static assertion so that later their
+    // argument expression can be replaced with the value to which the argument
+    // expression would notmally evaluate.
+    static_assertions_to_rewrite_.push_back(static_assert_decl);
```
afd, line 127: Before the change there was `return true`, which ensures that the expression associated with the static cast won't be recursed into.

You have replaced this with the logging that this rewrite is needed - but shouldn't both happen - i.e., logging, and then still doing a return? I think that with your current change the expression of the static assertion will be mutated, but then over-written. Could you investigate?

src/libdredd/src/mutate_visitor.cc
```diff
@@ -108,14 +108,23 @@ bool MutateVisitor::TraverseDecl(clang::Decl* decl) {
...
     // It does not make sense to mutate static assertions, as (a) this will
     // very likely lead to compile-time failures due to the assertion not
     // holding, (b) if compilation succeeds then the assertion is not actually
     // present at runtime so there is no notion of killing the mutant, and (c)
     // dynamic instrumentation of the mutation operator will break the rules
     // associated with static assertions anyway.
-    return true;
+
+    // However, some C++ compiler complain if the expression under static
+    // assertion is not a compile-time constants. To allow the constant
```
afd, line 121: I'd rephrase the first sentence to: "However, the expression under a static assertion must be a compile-time constant". I think this is a strict requirement (because the assertion must be evaluated at compile time) - I don't think this is something that only certain compilers require.

src/libdredd/src/mutate_visitor.cc
```diff
@@ -108,14 +108,23 @@ bool MutateVisitor::TraverseDecl(clang::Decl* decl) {
...
+    // declaration that occurs inside static assertion to have their initial
+    // value mutated (because they may be used elsewhere), but to avoid
+    // compilaton errors, we record the static assertion so that later their
+    // argument expression can be replaced with the value to which the argument
+    // expression would notmally evaluate.
```
afd, line 126: notmally -> normally. However, I think you noted in a comment on the issue that actually this gets replaced with 1 later, so can you update this comment.

src/libdredd/src/mutate_visitor.cc
```diff
@@ -108,14 +108,23 @@ bool MutateVisitor::TraverseDecl(clang::Decl* decl) {
...
     // dynamic instrumentation of the mutation operator will break the rules
     // associated with static assertions anyway.
-    return true;
+
+    // However, some C++ compiler complain if the expression under static
+    // assertion is not a compile-time constants. To allow the constant
+    // declaration that occurs inside static assertion to have their initial
+    // value mutated (because they may be used elsewhere), but to avoid
+    // compilaton errors, we record the static assertion so that later their
+    // argument expression can be replaced with the value to which the argument
```
afd, line 125: Careful with plurals - I'd say "we record static assertions so that later their argument expressions can be"

src/libdredd/src/mutate_visitor.cc
```diff
@@ -151,6 +160,17 @@ bool MutateVisitor::TraverseDecl(clang::Decl* decl) {
       return true;
     }
   }
+
+  // Beside variable declaration, constant-sized C++ array can occurs inside a
+  // struct as a field. As in the case of variable declaration, we record the
```
afd, line 165: Minor language nits:

Beside variable declaration -> Besides variable declarations

constant-sized C++ array can occurs -> constant-sized C++ arrays can occur

As in the case of variable declaration -> As in the case of variable declarations

afd (changes requested):
This looks good, but there's one functional change I am slightly unsure about and would like to re-review.

A couple of language nits to be fixed too.

src/libdredd/src/mutate_visitor.cc
JonathanFoo0523: changed

src/libdredd/src/mutate_visitor.cc
JonathanFoo0523: fixed

src/libdredd/src/mutate_visitor.cc
JonathanFoo0523: fixed

src/libdredd/src/mutate_visitor.cc
JonathanFoo0523: You are right that not having return here caused the expression under it to be recursed into. However, these expressions are not mutated because `MutateVisitor::VisitExpr` returns early due to these expressions not being `IsInFunction()`.

Now, I am not sure why `MutateVisitor::IsInFunction()` considers these expressions not inside a function. This could point to a bug in `MutateVisitor::IsInFunction()`.

JonathanFoo0523: The AST is as below:
```
-FunctionDecl 0x555558e59920 </home/ubuntu/dredd/examples/simple/assert.cc:1:1, line:3:1> line:1:5 main 'int ()'
  `-CompoundStmt 0x555558e59af0 <col:12, line:3:1>
    `-DeclStmt 0x555558e59ad8 <line:2:3, col:19>
      `-StaticAssertDecl 0x555558e59a98 <col:3, col:18> col:3
        |-ImplicitCastExpr 0x555558e59a60 <col:17> 'bool' <IntegralToBoolean>
        | `-IntegerLiteral 0x555558e59a40 <col:17> 'int' 1
        `-<<<NULL>>>
```
Note that static assertion is a declaration. 

If the program doesn't return at that point, the it will proceed with adding `StaticAssertDecl` to global variable `enclosing_decls_`. 

When the integer literal is traversed by  `MutateVisitor::VisitExpr`, it will check whether this expression is in a function(`MutateVisitor::IsInFunction()`). `MutateVisitor::IsInFunction()` loop through each enclosing declaration and check
(1) if the enclosing decl is a function declaration, return true
(2) if it is not a variable declaration, return false
(3) otherwise, continue looping 
Thus, the integer literal is "not in function" because the enclosing scope - `StaticAssertDecl` - is not a variable declaration.

afd: Interesting. I wonder then why I had the previous "return true" - it seems like I wouldn't have even needed the check for a static assertion given what you have discovered about `IsInFunction`. But also it seems unlikely that I would have added this for no reason.

JonathanFoo0523: Seems like there's no `MutateVisitor::IsInFunction()` back when the check for a static assertion is first written. The enclosing declaration is used in a simpler way. 
```
if (enclosing_decls_.empty()) {
    // Only consider mutating binary expressions that occur inside functions.
    return true;
  }
```

src/libdredd/src/mutate_ast_consumer.cc
```diff
@@ -119,20 +121,32 @@ void MutateAstConsumer::HandleTranslationUnit(clang::ASTContext& ast_context) {
...
+  // Rewrite the argument of static assertion as needed to `1`.
+  // There's no need to evaluate the actual expression, as any value other than
+  // 1 would have caused the front-end used by Dredd to fail.
+  for (const auto& static_assert_decl :
+       visitor_->GetStaticAssertionsToRewrite()) {
+    RewriteExpressionInMainFileToIntegral(static_assert_decl->getAssertExpr(),
+                                          1);
+  }
+
+  // Rewrite the constant integer argument of some function.
```
afd, line 141: -> "Rewrite the constant integer arguments of some builtin functions."

(Is what I propose accurate?)

src/libdredd/include_private/include/libdredd/mutate_visitor.h
```diff
@@ -103,11 +104,24 @@ class MutateVisitor : public clang::RecursiveASTVisitor<MutateVisitor> {
...
   }
 
+  // Yields the static assertions, whose expression need to be
+  // rewritten.
+  [[nodiscard]] const std::vector<const clang::StaticAssertDecl*>&
+  GetStaticAssertionsToRewrite() const {
+    return static_assertions_to_rewrite_;
+  }
+
+  // Yields the constant function arguments which need to be rewritten.
```
afd, line 119: Are these always for builtin functions? If so, could you change the name of the field and the getter to reflect that? E.g. `constant_builtin_function_arguments_to_rewrite` for the field, and `GetConstantBuiltinFunctionArgumentsToRewrite` for the getter. (I am OK with long names, as you have probably noticed!)

src/libdredd/src/mutate_visitor.cc
```diff
@@ -214,6 +233,20 @@ bool MutateVisitor::TraverseStmt(clang::Stmt* stmt) {
     }
   }
 
+  // Some functions, e.g., __builtin_frame_address(), require their arguments to
+  // be constant integers. Attempting to apply a mutator function to such an
+  // argument would result in a compile-time error. Thus, we record these
+  // arguments' expressions so that they can be rewritten with the values to
+  // which the expressions would normally evaluate.
+  if (const auto* call_expr = llvm::dyn_cast<clang::CallExpr>(stmt)) {
+    const unsigned int frame_address_builtin_id = 455;
```
afd, line 242: How did you work out this was 455? Are you sure there isn't a named constant in some header file which yields this value? If so, it would be much better to use the named constant.

I did some digging and it seems that if you do:

`#include "clang/TargetBuiltins.h"`

then you should be able to use:

`clang::Builtin::BI__builtin_frame_address`

I presume that will evaluate to the required constant - can you check please?

src/libdredd/src/mutate_visitor.cc
```diff
@@ -214,6 +233,20 @@ bool MutateVisitor::TraverseStmt(clang::Stmt* stmt) {
...
+    if (call_expr->getBuiltinCallee() == frame_address_builtin_id) {
```
afd, line 243: ... then you can use use the constant here and get rid of the `frame_address_builtin_id` temporary.

src/libdredd/src/mutate_visitor.cc
afd: Thanks for investigating the revision history here - and what you have done looks good.

afd (changes requested):
This is looking very good! Various nits, and one slightly more substantial change please.

src/libdredd/include_private/include/libdredd/mutate_ast_consumer.h
```diff
@@ -15,12 +15,15 @@
 #ifndef LIBDREDD_MUTATE_AST_CONSUMER_H
 #define LIBDREDD_MUTATE_AST_CONSUMER_H
 
+#include <stdint.h>
```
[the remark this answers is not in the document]
JonathanFoo0523: changed

src/libdredd/include_private/include/libdredd/mutate_ast_consumer.h
```diff
@@ -63,6 +66,9 @@ class MutateAstConsumer : public clang::ASTConsumer {
   [[nodiscard]] std::string GetMutantTrackingDreddPreludeC(
       int initial_mutation_id) const;
 
+  bool RewriteExpressionInMainFileToIntegral(const clang::Expr* expr,
+                                             uint64_t integral);
```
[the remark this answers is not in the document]
JonathanFoo0523: changed

src/libdredd/include_private/include/libdredd/mutate_visitor.h
JonathanFoo0523: changed

src/libdredd/include_private/include/libdredd/mutate_visitor.h
```diff
@@ -217,7 +231,15 @@ class MutateVisitor : public clang::RecursiveASTVisitor<MutateVisitor> {
 
   // This records C++ constant-sized array declarations, so that size
   // expressions can be rewritten with the integers to which they evaluate.
-  std::vector<const clang::VarDecl*> constant_sized_arrays_to_rewrite_;
+  std::vector<const clang::DeclaratorDecl*> constant_sized_arrays_to_rewrite_;
+
+  // This records static assertion declarations, so that its argument
```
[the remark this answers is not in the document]
JonathanFoo0523: changed

src/libdredd/include_private/include/libdredd/mutate_visitor.h
```diff
@@ -217,7 +231,15 @@ class MutateVisitor : public clang::RecursiveASTVisitor<MutateVisitor> {
...
+  // expressions can be rewritten with the integers to which they evaluate.
+  std::vector<const clang::StaticAssertDecl*> static_assertions_to_rewrite_;
+
+  // This records constant integer function argument, so that the expression can
```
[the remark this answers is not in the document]
JonathanFoo0523: changed

src/libdredd/src/mutate_visitor.cc
JonathanFoo0523: By running a program with __buitlin_frame_address and printing out the builtin id. Thank you for the suggestion, I didn't managed to find the named constant before.

changed to "#include "clang/Basic/TargetBuiltins.h", and rewrite to avoid the const variable declaration

src/libdredd/src/mutate_visitor.cc
JonathanFoo0523: done

src/libdredd/src/mutate_ast_consumer.cc
JonathanFoo0523: Correct. I was considering the possibility of some other non-builtin functions that require constant integer argument, thus use more general naming(`constant_arguments_to_rewrite_.push_back` instead of  `constant_builtin_function_arguments_to_rewrite_`). 

For now, I will assume only "some builtin_function" need to have their argument rewritten. We can change the variable name and comment if we discover this is not the case in the future.

afd:
After #300 lands, can you rebase this over main and re-generate expectations for single file test (as they may have changed as a result of #220)?

Then let me know when ready for a re-review.

afd (approved):

PR #297, opened 2024-07-16
Avoid redundant/equivalent unsigned comparison mutations

### LINKED ISSUE #240: Avoid equivalent mutation to unsigned int
Currently, for unsigned integer `a`, Dredd allow the mutation:

```
a != 0
```

to

```
a > 0
```

Both expressions are equivalent. Since Dredd has the type information when traversing the AST, it should be possible to disallow such mutation.

afd:
Fixes #240.


JamesLee-Jones (approved):
Looks good! It may be worth copying the test case added to C++ as we have done elsewhere in case the functionality diverges in the future (although unlikely).

test/single_file/unsigned_comparisons_with_zero.c.expected
```diff
@@ -0,0 +1,194 @@
...
+static int __dredd_replace_binary_operator_GE_arg1_unsigned_int_arg2_unsigned_int_lhs_zero(unsigned int arg1, unsigned int arg2, int local_mutation_id) {
+  if (!__dredd_some_mutation_enabled) return arg1 >= arg2;
+  if (__dredd_enabled_mutation(local_mutation_id + 0)) return arg1 == arg2;
+  if (__dredd_enabled_mutation(local_mutation_id + 1)) return arg1 > arg2;
+  return arg1 >= arg2;
+}
+
+static int __dredd_replace_binary_operator_EQ_arg1_unsigned_int_arg2_unsigned_int_rhs_zero(unsigned int arg1, unsigned int arg2, int local_mutation_id) {
+  if (!__dredd_some_mutation_enabled) return arg1 == arg2;
+  if (__dredd_enabled_mutation(local_mutation_id + 0)) return arg1 >= arg2;
```
JonathanFoo0523, line 128: Mutation of a == 0

a >= 0 is subsumed by true

JonathanFoo0523 (commented):
I think the role of candidate mutation and the binary operator to apply mutation ion is switched on `IsRedundantReplacementForUnsignedComparison`, and as a result, subsumed mutation are produced. I might be wrong tho. Let me know if I am wrong, I will have another look.

Does it work if we don't have an unsigned literal(eg: `x !=0` instead of `x != 0u`)?

afd:
@JamesLee-Jones Thanks for the comment. The C++ single file tests take considerably longer to run compared with the C ones, so I have decided not to add them in this instance, feeling confident this won't change.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -1050,6 +1056,65 @@ bool MutationReplaceBinaryOperator::
...
 
+bool MutationReplaceBinaryOperator::IsRedundantReplacementForUnsignedComparison(
+    clang::BinaryOperatorKind operator_kind,
+    const clang::ASTContext& ast_context) const {
+  if (binary_operator_->getLHS()->getType()->isUnsignedIntegerType() &&
+      binary_operator_->getRHS()->getType()->isUnsignedIntegerType()) {
+    if (MutationReplaceExpr::ExprIsEquivalentToInt(*binary_operator_->getLHS(),
+                                                   0, ast_context)) {
+      // LHS is 0
+      switch (binary_operator_->getOpcode()) {
```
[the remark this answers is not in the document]
afd: You're right! Fix coming up.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -1050,6 +1056,65 @@ bool MutationReplaceBinaryOperator::
...
+    }
+    if (MutationReplaceExpr::ExprIsEquivalentToInt(*binary_operator_->getRHS(),
+                                                   0, ast_context)) {
+      // RHS is 0
+      switch (binary_operator_->getOpcode()) {
+        case clang::BO_LT:
+        case clang::BO_GE:
+          // "a < 0" is subsumed by "false"
+          // "a >= 0" is subsumed by "true"
+          return true;
```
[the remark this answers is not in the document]
afd: Agreed - fix coming up.

test/single_file/unsigned_comparisons_with_zero.c.expected
afd: Re-generated tests, according to fix, coming up.

test/single_file/unsigned_comparisons_with_zero.c.expected
```diff
@@ -0,0 +1,194 @@
...
+  if (__dredd_enabled_mutation(local_mutation_id + 0)) return !(arg);
+  if (__dredd_enabled_mutation(local_mutation_id + 1)) return ~(arg);
+  if (__dredd_enabled_mutation(local_mutation_id + 2)) return 0;
+  if (__dredd_enabled_mutation(local_mutation_id + 3)) return 1;
+  return arg;
+}
+
+static int __dredd_replace_binary_operator_NE_arg1_unsigned_int_arg2_unsigned_int_rhs_zero(unsigned int arg1, unsigned int arg2, int local_mutation_id) {
+  if (!__dredd_some_mutation_enabled) return arg1 != arg2;
+  if (__dredd_enabled_mutation(local_mutation_id + 0)) return arg1 < arg2;
```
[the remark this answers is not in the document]
afd: Re-generated tests, according to fix, coming up.

afd:
@JonathanFoo0523 - the optimisation does work if 0u is replaced by 0 - in next iteration I have added tests to check for this.

It raises room for another optimisation, though, which is that when an int is cast (implicitly or otherwise) to unsigned, we mutate both before and after the cast, which will lead to replacement with 0 and 1 occurring redundantly as they will be used in both places. I'll open an issue for this.

afd:
@JonathanFoo0523 Thanks for your comments - good spots. I have revised - and to make it less likely that I'll make a mistake like this again I renamed "operator_kind" to "replacement_operator" in all of the functions that are considering whether a replacement is legitimate (both for unary and binary operator mutations).

afd:
> It raises room for another optimisation, though, which is that when an int is cast (implicitly or otherwise) to unsigned, we mutate both before and after the cast, which will lead to replacement with 0 and 1 occurring redundantly as they will be used in both places. I'll open an issue for this.

Actually I think I got this wrong - perhaps I looked at the non-optimised code.

test/single_file/unsigned_comparisons_with_zero.c.expected
```diff
@@ -0,0 +1,246 @@
...
+
+static int __dredd_replace_binary_operator_LT_arg1_unsigned_int_arg2_unsigned_int_lhs_zero(unsigned int arg1, unsigned int arg2, int local_mutation_id) {
+  if (!__dredd_some_mutation_enabled) return arg1 < arg2;
+  if (__dredd_enabled_mutation(local_mutation_id + 0)) return arg1 != arg2;
+  return arg1 < arg2;
+}
+
+static int __dredd_replace_binary_operator_LE_arg1_unsigned_int_arg2_unsigned_int_rhs_zero(unsigned int arg1, unsigned int arg2, int local_mutation_id) {
+  if (!__dredd_some_mutation_enabled) return arg1 <= arg2;
+  if (__dredd_enabled_mutation(local_mutation_id + 0)) return arg1 == arg2;
```
JonathanFoo0523, line 91: Something might not be right here, as

a <= 0 is equivalent to a == 0

test/single_file/unsigned_comparisons_with_zero.c.expected
```diff
@@ -0,0 +1,246 @@
...
+static int __dredd_replace_binary_operator_LE_arg1_unsigned_int_arg2_unsigned_int_lhs_zero(unsigned int arg1, unsigned int arg2, int local_mutation_id) {
+  if (!__dredd_some_mutation_enabled) return arg1 <= arg2;
+  if (__dredd_enabled_mutation(local_mutation_id + 0)) return arg1 == arg2;
+  if (__dredd_enabled_mutation(local_mutation_id + 1)) return arg1 < arg2;
+  return arg1 <= arg2;
+}
+
+static int __dredd_replace_binary_operator_GT_arg1_unsigned_int_arg2_unsigned_int_rhs_zero(unsigned int arg1, unsigned int arg2, int local_mutation_id) {
+  if (!__dredd_some_mutation_enabled) return arg1 > arg2;
+  if (__dredd_enabled_mutation(local_mutation_id + 0)) return arg1 != arg2;
```
JonathanFoo0523, line 104: something wrong

a > 0 is equivalent to a != 0

JonathanFoo0523 (commented):
Didn't look through the source code as I think the produced expected testcase doesn't look right. I will re-review the code after.

test/single_file/unsigned_comparisons_with_zero.c.expected
afd: Good catch - I have made the equivalence check symmetric now.

test/single_file/unsigned_comparisons_with_zero.c.expected
afd: Good catch - I have made the equivalence check symmetric now.

test/single_file/unsigned_comparisons_with_zero.c.expected
```diff
@@ -0,0 +1,246 @@
...
+
+static int __dredd_replace_binary_operator_NE_arg1_unsigned_int_arg2_unsigned_int_lhs_zero(unsigned int arg1, unsigned int arg2, int local_mutation_id) {
+  if (!__dredd_some_mutation_enabled) return arg1 != arg2;
+  return arg1 != arg2;
+}
+
+static int __dredd_replace_binary_operator_LT_arg1_unsigned_int_arg2_unsigned_int_rhs_zero(unsigned int arg1, unsigned int arg2, int local_mutation_id) {
+  if (!__dredd_some_mutation_enabled) return arg1 < arg2;
+  if (__dredd_enabled_mutation(local_mutation_id + 0)) return arg1 != arg2;
+  if (__dredd_enabled_mutation(local_mutation_id + 1)) return arg1 <= arg2;
```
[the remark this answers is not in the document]
afd: I have opened #303 to track this - it's a good point but is separate from the issue being addressed by this PR.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -1024,34 +1029,108 @@ MutationReplaceBinaryOperator::ClangOperatorKindToProtobufOperatorKind(
...
+      //
+      // These are subsumed by replacing the overall expression with "false"
+      // and "true", respectively.
+      return true;
+    }
+    if (std::set({binary_operator_->getOpcode(), replacement_operator}) ==
+        std::set<clang::BinaryOperatorKind>({clang::BO_EQ, clang::BO_GE})) {
+      // "0 == a" is equivalent to "0 >= a"
+      return true;
+    }
```
JamesLee-Jones, line 1096: By the fact that `0 == a` is equivalent to `0 >= a` and commutativity, it is also redundant to replace `0 >= a` by `0 == a` for signed integers (which is unlikely but may occur). I think equality on ordered sets compare element wise so this wouldn't currently cover the second case.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -1024,34 +1029,108 @@ MutationReplaceBinaryOperator::ClangOperatorKindToProtobufOperatorKind(
...
+    if (std::set({binary_operator_->getOpcode(), replacement_operator}) ==
+        std::set<clang::BinaryOperatorKind>({clang::BO_NE, clang::BO_LT})) {
+      // "0 != a" equivalent to "0 < a"
+      return true;
+    }
+  }
```
JamesLee-Jones, line 1102: As above, it is also redundant to replace `0 < a` with `0 != a`

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -1024,34 +1029,108 @@ MutationReplaceBinaryOperator::ClangOperatorKindToProtobufOperatorKind(
...
+      //
+      // These are subsumed by replacing the overall expression with "false"
+      // and "true", respectively.
+      return true;
+    }
+    if (binary_operator_->getOpcode() == clang::BO_EQ &&
+        replacement_operator == clang::BO_LE) {
+      // "a == 0" is equivalent to "a <= 0"
+      return true;
+    }
```
JamesLee-Jones, line 1122: As above, it is also redundant to replace `a == 0` with `0 != a`.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -1024,34 +1029,108 @@ MutationReplaceBinaryOperator::ClangOperatorKindToProtobufOperatorKind(
...
+    if (binary_operator_->getOpcode() == clang::BO_NE &&
+        replacement_operator == clang::BO_GT) {
+      // "a != 0" equivalent to "a > 0"
+      return true;
```
JamesLee-Jones, line 1126: As above, it is also redundant to replace `a != 0` with `0 > a`.

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -1024,34 +1029,108 @@ MutationReplaceBinaryOperator::ClangOperatorKindToProtobufOperatorKind(
...
+      // one of:
+      //
+      // - "a < 0"
+      // - "a >= 0"
+      //
+      // These are subsumed by replacing the overall expression with "false"
+      // and "true", respectively.
+      return true;
+    }
+    if (binary_operator_->getOpcode() == clang::BO_EQ &&
```
JamesLee-Jones, line 1118: Above you used set equality but here used two equality checks, is there any reason?

src/libdredd/src/mutation_replace_binary_operator.cc
```diff
@@ -1024,34 +1029,108 @@ MutationReplaceBinaryOperator::ClangOperatorKindToProtobufOperatorKind(
...
+        replacement_operator == clang::BO_LE) {
+      // "a == 0" is equivalent to "a <= 0"
+      return true;
+    }
+    if (binary_operator_->getOpcode() == clang::BO_NE &&
+        replacement_operator == clang::BO_GT) {
```
JamesLee-Jones, line 1124: Above you used set equality but here used two equality checks, is there any reason?

JamesLee-Jones (changes requested):
I believe there are a couple of cases that have been missed and a couple of small questions but otherwise looks good!

test/single_file/unsigned_comparisons_with_zero.c.expected
```diff
@@ -0,0 +1,244 @@
...
+  return arg1 < arg2;
+}
+
+static int __dredd_replace_binary_operator_LT_arg1_unsigned_int_arg2_unsigned_int_lhs_zero(unsigned int arg1, unsigned int arg2, int local_mutation_id) {
+  if (!__dredd_some_mutation_enabled) return arg1 < arg2;
+  return arg1 < arg2;
```
JonathanFoo0523, line 89:
```
+----+-----+------+-------+
|  a | 0<a | TRUE | FALSE |
+----+-----+------+-------+
|  0 |  F  |   T  |   F   |
+----+-----+------+-------+
| >0 |  T  |   T  |   F   |
+----+-----+------+-------+
```
correct

test/single_file/unsigned_comparisons_with_zero.c.expected
```diff
@@ -0,0 +1,244 @@
...
+static int __dredd_replace_binary_operator_NE_arg1_unsigned_int_arg2_unsigned_int_lhs_zero(unsigned int arg1, unsigned int arg2, int local_mutation_id) {
+  if (!__dredd_some_mutation_enabled) return arg1 != arg2;
+  return arg1 != arg2;
+}
+
+static int __dredd_replace_binary_operator_LT_arg1_unsigned_int_arg2_unsigned_int_rhs_zero(unsigned int arg1, unsigned int arg2, int local_mutation_id) {
+  if (!__dredd_some_mutation_enabled) return arg1 < arg2;
+  if (__dredd_enabled_mutation(local_mutation_id + 0)) return arg1 != arg2;
+  if (__dredd_enabled_mutation(local_mutation_id + 1)) return arg1 <= arg2;
+  return arg1 < arg2;
```
JonathanFoo0523, line 84:
```
+----+-----+------+------+
|  a | a<0 | a<=0 | a!=0 |
+----+-----+------+------+
|  0 |  F  |   T  |   F  |
+----+-----+------+------+
| >0 |  F  |   F  |   T  |
+----+-----+------+------+
```
correct

test/single_file/unsigned_comparisons_with_zero.c.expected
```diff
@@ -0,0 +1,244 @@
...
+}
+
+static int __dredd_replace_binary_operator_NE_arg1_unsigned_int_arg2_unsigned_int_rhs_zero(unsigned int arg1, unsigned int arg2, int local_mutation_id) {
+  if (!__dredd_some_mutation_enabled) return arg1 != arg2;
+  return arg1 != arg2;
+}
+
+static int __dredd_replace_binary_operator_NE_arg1_unsigned_int_arg2_unsigned_int_lhs_zero(unsigned int arg1, unsigned int arg2, int local_mutation_id) {
+  if (!__dredd_some_mutation_enabled) return arg1 != arg2;
+  return arg1 != arg2;
```
JonathanFoo0523, line 77: same as above

correct

PR #305, opened 2024-07-18
Avoid mutation of expression under `Decltype`

### LINKED ISSUE #304: Dredd introduces lambda in unevaluated expression
Applying dredd and compiling the following program
```
int bar() {}

void foo() {
  unsigned E = (decltype(bar()))1000;
}
```
mutate the cast expression to
```
(decltype(__dredd_replace_expr_int([&]() -> int { return static_cast<int>(bar()); }, 0)))1000
```
and results in
```
error: lambda expression in an unevaluated operand
```

JonathanFoo0523:
Since Dredd's mutator functions don't change the type of expression,
mutating an expression under `Decltype` won't affect the outcome of
casting (i.e., they introduce equivalent mutation). Furthermore,
mutation of an expression under `Decltype` could introduce lambda
expression in the unevaluated operand.

Fixes #304

### DIFF
--- a/src/libdredd/src/mutate_visitor.cc
+++ b/src/libdredd/src/mutate_visitor.cc
@@ -265,6 +265,14 @@ bool MutateVisitor::TraverseStmt(clang::Stmt* stmt) {
     }
   }
 
+  // Do not mutate expression under `decltype`, as Dredd doesn't change the type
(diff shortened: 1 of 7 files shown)


afd (approved):
LGTM.

PR #306, opened 2024-07-19
Avoid static cast bypass for parenthesis expression in initializer list

### LINKED ISSUE #301: Parenthesis expression bypass `static_cast` insertion in initializer list
Applying dredd to
```
struct foo {
  short x;
  short y;
};

void baz() {
  foo bar = {1, (2)};
}
```
mutate the initializer list to
```
{static_cast<short>(__dredd_replace_expr_int_one(1, 0)), (__dredd_replace_expr_int_constant(2, 3))};
```
which results in compilation error:
```
error: non-constant-expression cannot be narrowed from type 'int' to 'short' in initializer list
```

JonathanFoo0523:
Fix a problem where expressions in parentheses within an initializer
list bypass being explicitly cast with `static_cast`, causing issues
when the expression is implicitly cast to a narrower type.

Fixes #301

### DIFF
--- a/src/libdredd/src/mutation_replace_expr.cc
+++ b/src/libdredd/src/mutation_replace_expr.cc
@@ -584,7 +584,7 @@ void MutationReplaceExpr::ReplaceExprWithFunctionCall(
   // cast.
   // For the following check to work, we need to move up any enclosing
   // parentheses.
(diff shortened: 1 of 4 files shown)


afd (approved):
LGTM!

PR #332, opened 2024-09-15
Rewrite const expr used as template arguments

### LINKED ISSUE #331: dredd mutates constant expression used as template argument
Applying dredd to
```
template<typename T, int N>
class auto_vec {};

void foo() {
    const int alloc_count = 42;
    auto_vec<int, alloc_count> a;
}
```
results in compilation error
```
error: non-type template argument is not a constant expression
    auto_vec<int, alloc_count> a;
                  ^~~~~~~~~~~
```
due to the mutation
```
const int alloc_count = __dredd_replace_expr_int_constant(42, 0);
```

JonathanFoo0523:
Prevents an issue where constant expressions used in template arguments
could have their initializers altered, causing the template argument to
no longer be a compile-time constant. We rewrite the template argument
expressions to substitute them with their evaluated values. This allows
the constants within the template argument expressions — potentially
referenced elsewhere — to have their initial values modified freely.

Fixes #331 .


src/libdredd/src/mutate_visitor.cc
```diff
@@ -859,4 +878,33 @@ bool MutateVisitor::IsLvalueCallThatUsesMaterializedTemporary(
...
 }
 
+void MutateVisitor::SaveConstantTemplateArgumentForRewrite(
+    const clang::TemplateSpecializationType* template_specialization_type) {
+  for (auto template_argument :
+       template_specialization_type->template_arguments()) {
+    switch (template_argument.getKind()) {
+      case clang::TemplateArgument::Expression:
+        // Template Argument is Expression
+        if (const auto* expr = template_argument.getAsExpr()) {
```
afd, line 888: Is it possible for getKind() to return Expression but for getAsExpr() to then fail? Likely not, therefore I would get rid of the if and do something like:

```
const auto* expr = ...;
assert(expr != nullptr && "It should be possible to get a template argument with Expression kind as an Expr");
constant_arguments_to_rewritte.push_back(expr);
```

afd (approved):
This looks good to me - thanks. Please see a couple of small changes to attend to in my review, but otherwise good to merge.

src/libdredd/src/mutate_visitor.cc
JonathanFoo0523: Is it okay to replace with
```
constant_arguments_to_rewritte.push_back(template_argument.getAsExpr());
```
instead, without the assertion checking?

src/libdredd/include_private/include/libdredd/mutate_visitor.h
```diff
@@ -309,10 +316,10 @@ class MutateVisitor : public clang::RecursiveASTVisitor<MutateVisitor> {
   // expressions can be rewritten with the integers to which they evaluate.
   std::vector<const clang::StaticAssertDecl*> static_assertions_to_rewrite_;
 
-  // This records constant integer function arguments, so that their expressions
-  // can be rewritten with the integers to which they evaluate.
-  std::vector<const clang::Expr*>
-      constant_builtin_function_arguments_to_rewrite_;
+  // This records constant integer builtin_functions' and templates' arguments,
```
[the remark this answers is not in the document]
JonathanFoo0523: done

src/libdredd/src/mutate_visitor.cc
afd: Yes, please go ahead. Once done, if bots pass, feel free to merge.

PR #336, opened 2024-09-16
Insert prelude at start of source file by default

### LINKED ISSUE #333: Insert prelude at beginning of file to minimise manual modifications
The current prelude-injection behavior leads to many files requiring manual modification with:

```cpp
void __dredd_prelude_start();
```

Almost all source files with `override` class function modifiers fail without this manual modification, which is quite common. 

I believe we could reduce the need for manual changes by inserting the prelude at the very beginning of the file. Only a few cases, such as #302, wouldn't work with this approach, but I think those cases are rare.

JonathanFoo0523:
Dredd previously inserted the prelude before the first function to avoid
placing it before feature test macros or headers that must precede other
headers. However, this heuristic often fails in files with class methods
using the `override` modifier, which is common. We now insert the
prelude at the start of the file. In rare cases where the prelude must
follow feature test macros or specific headers, manual insertion of
`void __dredd_prelude_start();` will be required.

Fixes #333 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,8 +382,8 @@ this feature in your project's build configuration.
 
 The mutations that Dredd applies rely on a number of function definitions that must be present at a suitable point in each file that Dredd mutates.
 These are referred to as the Dredd *prelude*.
-Dredd uses a heuristic to determine where to place its prelude.
-Sometimes this heuristic may not find an appropriate place, e.g. because of limitations or bugs in Clang's libTooling (see [this issue](https://github.com/mc-imperial/dredd/issues/322) for example).
+Dredd inserts prelude at the beginning of source files that may produce mutants.
+Sometimes this heuristic may not be appropriate, e.g. because a feature test macro must be defined before including any header files (see [this issue](https://github.com/mc-imperial/dredd/issues/215) for example).
 
 If you face a problem where Dredd-mutated code does not compile due to the Dredd prelude being injected in a bad place, you can declare a function prototype with void return type and no arguments called `__dredd_prelude_start` somewhere in your source file.
 Dredd will then insert its prelude right before this function prototype (regardless of whether this is a sensible place to insert the prelude).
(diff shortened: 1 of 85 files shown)


README.md
```diff
@@ -382,8 +382,8 @@ this feature in your project's build configuration.
 
 The mutations that Dredd applies rely on a number of function definitions that must be present at a suitable point in each file that Dredd mutates.
 These are referred to as the Dredd *prelude*.
-Dredd uses a heuristic to determine where to place its prelude.
-Sometimes this heuristic may not find an appropriate place, e.g. because of limitations or bugs in Clang's libTooling (see [this issue](https://github.com/mc-imperial/dredd/issues/322) for example).
+Dredd inserts prelude at the beginning of source files that may produce mutants.
```
afd, line 385: "inserts prelude" -> "inserts its prelude"

"at the beginning of source files that may produce mutants" -> "at the beginning of each mutated source file"

src/libdredd/src/mutate_visitor.cc
```diff
@@ -734,7 +692,6 @@ bool MutateVisitor::TraverseCompoundStmt(clang::CompoundStmt* compound_stmt) {
 }
 
 void MutateVisitor::AddMutation(std::unique_ptr<Mutation> mutation) {
-  UpdateStartLocationOfFirstFunctionInSourceFile();
```
afd, line 737: We introduced the AddMutation function due to the need to call UpdateStartLocation... before each addition of a mutation. I wondered therefore if we should revert this and get rid of the helper, and outline the mutation_tree_path.back()... statements. However, on balance I think it's nice to encapsulate the management of the mutation tree into this function so let's keep it. (So, no action - feel free to resolve, just FYI.)

afd (approved):
LGTM - please attend to the minor change and then feel free to merge.

afd:
@JamesLee-Jones FYI - we are reverting to putting the prelude at the start of the file, but see a recent change that introduce a new `__dredd_prelude_start()` placeholder function that can be used to specify a different location for the prelude (useful if feature test macros are present).

afd:
Due to #337 this will need a rebase and force-push, and the expectation files will need to be regenerated.

README.md
JonathanFoo0523: done

PR #343, opened 2024-09-20
Use same set of headers in default mode and mutant-tracking mode

### LINKED ISSUE #342: Header consistency in dredd's default and mutant-tracking modes
Dredd uses different sets of headers for default mode and mutant-tracking mode. Since Dredd's headers are inserted at the beginning of the source file, they can "hide" the original program's headers and alter the definition of certain macros as described [here](https://github.com/mc-imperial/dredd/issues/341). As a result, the macro definitions may differ between default mode and mutant-tracking mode. Additionally, there might be cases where `void __dredd_prelude_start();` is required in one mode but not in the other.

To ensure consistency, Dredd should emit the union of #include headers from both modes.

JonathanFoo0523:
Previously, dredd included different sets of headers for default mode
and mutant-tracking mode. This could alter macro definitions and lead to
the manual insertion of `void __dredd_prelude_start();` in one mode but
not the other. To ensure consistency across both modes, Dredd now emits
the union of headers from both modes.

Fixes #342 .


JonathanFoo0523:
@afd seems like the window build failed because `stdatomic.h` is an experimental feature for MSVC:
[https://developercommunity.visualstudio.com/t/VS2022-1754:-stdatomich:-Compile-erro/10351339](https://developercommunity.visualstudio.com/t/VS2022-1754:-stdatomich:-Compile-erro/10351339)

The workaround is adding `/experimental:c11atomics` flag. 

Could you let me know where to modify the workflow script to add MSVC compiler flag?

afd:
From the CI logs I see:

cl.exe '/std:c17 /experimental:c11atomics' -c add.c

It looks like the two flags are being passed as if they were a single flag, due to the use of quotes. Hopefully if the quotes are eliminated this will work.

.github/workflows/build.sh
```diff
@@ -247,7 +247,8 @@ case "$(uname)" in
   rm -r test/execute/constant_function_argument
   
   # The C code generated by Dredd may require recent C/C++ support.
-  export DREDD_EXTRA_C_ARGS="/std:c17"
+  # Dredd's prelude includes experimental `stdatomic.h`` header.
```
afd, line 250: Remove redundant second backtick after `stdatomic.h`.

Nit: add "the" before "experimental".

.github/workflows/build.sh
```diff
@@ -247,7 +247,8 @@ case "$(uname)" in
...
+  export DREDD_EXTRA_C_ARGS=/std:c17 /experimental:c11atomics
```
afd, line 251: Not quite sure what needs to be done to get this to work.

afd (changes requested):
Looks good but some changes suggested.

src/libdredd/src/mutate_ast_consumer.cc
```diff
@@ -241,9 +241,12 @@ std::string MutateAstConsumer::GetRegularDreddPreludeCpp(
       (num_mutations + kWordSize - 1) / kWordSize;
 
   std::stringstream result;
+  result << "#include <atomic>\n";
```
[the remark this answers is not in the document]
JonathanFoo0523: Done

.github/workflows/build.sh
JonathanFoo0523: Done

.github/workflows/build.sh
JonathanFoo0523: Fixed

